PR #41, opened 2014-10-16
Added SVG to PNG conversion with batik-rasterizer for mj-page

geyang:
Known concerns:

I realized that the way I extracted the common `defs` element from the top of the page and attached them to each SVG when rendering might not work if the common defs option is set to false. I might fix this later.


episodeyang:
Thanks!! Just saw this.

Issue #54, opened 2014-11-24
Progress notification callback


anandthakker:
For large documents with lots of math, this api appears distinctly slower than in the browser.  (I haven't profiled, but I'm guessing it's due to jsdom...)

This probably isn't a big problem, since this api likely isn't being used to render math on a page _while_ it's being read.  However, I'd find it useful to get percentage-complete callbacks.  

[The approach described here](https://groups.google.com/forum/#!msg/mathjax-users/yk0gj6UC-KM/If-6q-G581oJ) would work, except that there isn't currently a way to touch MathJax.Hub from outside the API.  

Would it make sense to add as a config option an onInitComplete() callback, which could be called at the end of AuthorInit()?  This could serve as an analogue to the ability in the browser to have arbitrary mathjax configuration code.

pkra:
In the light of #206 (dropping mj-page), this feature request seems to become void.

pkra:
Closing this as `mj-page` has been deprecated in v1.

PR #55, opened 2014-11-24
Add onInitComplete hook to configuration

anandthakker:
This PR just adds an initialization callback, allowing client code authors to then run arbitrary mathjax config code after the MathJax object is fully setup but before it queues anything.

See https://github.com/mathjax/MathJax-node/issues/54 for an example use case.


pkra:
And another one we should revisit; it needs fixing though and needs to aim at `develop`. It would be great to restart if you're interested!

PR #72, opened 2015-02-10
Changed mj-single.js and tex2* to allow invalid TeX command

sammarshallou:
Before this change, use of an unsupported TeX command e.g. \frog
causes an error. This change adds an option (default false) to
permit invalid commands by using the noUndefined.js MathJax
extension. This results in the command being displayed as literal
text, in red, in the resulting SVG output.

I added --allowinvalid command-line flag to the tex2* utilities
to use the new feature.

### DIFF
--- a/bin/tex2mml
+++ b/bin/tex2mml
@@ -51,6 +51,10 @@ var argv = require("yargs")
     notexhints: {
       boolean: true,
       describe: "don't add TeX-specific classes"
+    },
+    allowinvalid: {
+      boolean: true,
+      describe: "do not fail for invalid TeX commands (display red)"
     }
   })
   .argv;
@@ -61,7 +65,8 @@ mjAPI.config({
       semantics: argv.semantics,
       texHints: !argv.notexhints
     }
-  }
+  },
+  allowInvalidTeX: argv.allowinvalid
 });
 mjAPI.start();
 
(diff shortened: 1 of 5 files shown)


sammarshallou:
Not sure if this is wanted, but I need it for my project (we have to make a best attempt to render equations even if they have unknown junk in) so I thought I'd submit in case others want it.

pkra:
I like it. @dpvc what do you think?

dpvc:
I think it might be better to provide a more general means of including extensions.  Perhaps 

    --extensions="TeX/noUndefined,MathML/content-mathml"

or something of that sort.  If we make individual command-line options for each extension, things could get out of hand.

Also, now that we are using the git-flow model, the pull request would need to branch from `develop` and merge to `develop` (rather than from and to `master`).

sammarshallou:
Understood! I agree there might be a need for including other extensions. I'll investigate doing it as suggested, and update/resubmit this. (Also will make it start from the other branch.)

I have another project at the moment so there might be some days before I get around to doing this but I presume nobody's in a hurry. :)

thanks.

dpvc:
Thanks for being willing to make the changes.  We appreciate the support!

sammarshallou:
I am closing this request - superceded by #75, as per dpvc's suggestion.

PR #73, opened 2015-02-12
Use OS temp directory os.tmpdir() rather than hard-coded /tmp

sammarshallou:
This makes it easier to use on operating systems (Windows) where /tmp
normally does not exist (and is probably the right thing to do anyhow).

### DIFF
--- a/lib/mj-page.js
+++ b/lib/mj-page.js
@@ -32,6 +32,7 @@ var fmt = require('util').format;
 var jsdom = require("jsdom").jsdom;
 var exec = require('child_process').exec;
 var speech = require('speech-rule-engine');
+var os = require('os');
 
 var displayMessages = false;      // don't log Message.Set() calls
 var displayErrors = true;         // show error messages on the console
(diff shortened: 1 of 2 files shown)


dpvc:
This looks good.  Although I've done a lot of javascript programming, I am relatively new to node, and don't know all the libraries, so I appreciate the correction.

I suspect that Peter will ask that you sign a CLA before we merge your changes.  @pkra, is that correct?

pkra:
@sammarshallou could you go fill out http://www.mathjax.org/cla.html ? Thanks!

sammarshallou:
Sure, I've signed the form.

pkra:
@sammarshallou Thanks! Are you contributing privately or as an employee?

sammarshallou:
As an employee. Hope that's okay! I did fill in the relevant box.

pkra:
Not as such but we've actually never had that case before. I'll check what we need and follow up by email.

PR #74, opened 2015-02-12
Changed Batik calls to use execFile instead of exec

sammarshallou:
In addition to being more secure, execFile (with arguments as an array)
should work on all platforms because it does not require a hardcoded
quote character. (On windows, quoting with the single ' causes this
call to fail.)


sammarshallou:
Think this is probably the 'correct' fix - there might be a node.js version requirement for execFile to be available though? Not sure.

dpvc:
These changes look good as well, and I will be happy to merge them when @pkra confirms that.

PR #75, opened 2015-02-13
Allow extensions to be specified on command line

sammarshallou:
Adds an 'extensions' option to config which takes a comma-separated
list of extensions to add to the default set. Then adds --extensions
parameter to each command-line tool to specify this value.

This allows use of MathJax features such as displaying invalid TeX
instead of failing to render (--extensions TeX/noUndefined), and
stripping out dangerous links from MathML (--extensions Safe).

### DIFF
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -37,6 +37,7 @@ var speech = require('speech-rule-engine');
 var displayMessages = false;      // don't log Message.Set() calls
 var displayErrors = true;         // show error messages on the console
 var undefinedChar = false;        // unknown characters are not saved in the error array
+var extensions = '';              // no additional extensions used
 
 var defaults = {
   ex: 6,                          // ex-size in pixels
@@ -315,6 +316,16 @@ function ConfigureMathJax() {
     }
   };
 
+  if (extensions) {
+    //
+    // Parse added extensions list and add to standard ones
+    //
+    var extensionList = extensions.split(/s*,\s*/);
+    for (var i = 0; i < extensionList.length; i++) {
+      var matches = extensionList[i].match(/^(.*?)(\.js)?$/);
+      window.MathJax.extensions.push(matches[1] + '.js');
+    }
+  }
   if (MathJaxConfig) {Insert(window.MathJax,MathJaxConfig)}
 }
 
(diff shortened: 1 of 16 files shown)


sammarshallou:
This follows on from #72. Here are some examples of usage with the 'noUndefined' TeX extension:

    $ bin/tex2svg 'a^\frog'
    TeX parse error: Undefined control sequence \frog

    $ bin/tex2svg 'a^\frog' --extensions=TeX/noUndefined
    <svg xmlns:xlink="http://www.w3.org/1999/xlink" ...

Or with the 'Safe' extension, for example with MathML:

    $ bin/mml2svg '<math><mtext href="javascript:alert()">X</mtext></math>'
    ...
    <a class="mjx-svg-href" href="javascript:alert()">
    ...

    $ bin/mml2svg '<math><mtext href="javascript:alert()">X</mtext></math>' --extensions Safe
    ... (SVG code with no link) ...

lib/mj-single.js
```diff
@@ -315,6 +316,28 @@ function ConfigureMathJax() {
...
+  if (extensions) {
+    //
+    // Parse added extensions list and add to standard ones
+    //
+    var extensionList = extensions.split(/s*,\s*/);
+    for (var i = 0; i < extensionList.length; i++) {
+      var matches = extensionList[i].match(/^(?:(.+)\/)?(.*)$/);
+      var category = matches[1], name = matches[2];
+      if (!category) {
+        window.MathJax.extensions = window.MathJax.extensions.concat(name + '.js');
```
dpvc: This could be simplified as

    window.MathJax.extensions.push(name+'.js');

lib/mj-single.js
```diff
@@ -315,6 +316,28 @@ function ConfigureMathJax() {
...
+          window.MathJax[category] = {};
+        }
+        if (!window.MathJax[category].extensions) {
+          window.MathJax[category].extensions = [];
+        }
+        window.MathJax[category].extensions =
+          window.MathJax[category].extensions.concat(name + '.js');
+      }
+    }
+  }
```
dpvc: You co to a lot of effort here to break up the extensions into their individual categories, but that is not strictly necessary.  The main `extensions` array can include things like `"TeX/noUndefined.js"`, so it is OK to leave them all there.  The only reason to move them to the categories is that they would only be loaded when the input jax was loaded, but MathJax-node loads all three input jax automatically, so there is really no savings for that.  Might as well not introduce the added complexity of breaking up the list.

dpvc: You might want to consider allowing `--extensions "TeX/noUndefined.js"` as well as `--extensions "TeX/noUndefined"`.  In that case, you only want to add the `.js` if it isn't already there.

dpvc:
This looks good, but I did make a couple of comments in the code.

sammarshallou:
Thanks for comments! I've amended my commit to include the 3 requested changes:
1) array.push (don't understand why I wasn't doing that in the first place)
2) I hadn't realised you could specify the extensions at top level, that does make the code an awful lot simpler!
3) I made it allow .js

I retested it only using the examples I gave above (except I also tried adding .js this time), which is a bit less thorough than the testing I did last time but probably okay. Seems to still work...

PR #79, opened 2015-03-15
Separate entry points for single and page.

kintesh:
Provides two separate entry points for single and page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ batik/*
 !batik/README.md
 mathjax/*
 !mathjax/README.md
+
+# WebStorm
+.idea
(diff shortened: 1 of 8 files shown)


pkra:
I like the idea (though the PR should be to `develop`, not `master`). @dpvc what do you think?

kintesh:
My bad. delete this pull request, I'll make another one.

dpvc:
I'm not certain that there doesn't have to be more work done to keep the two versions separate.  They don't currently isolate their global variables, so it may be that loading the two causes conflicts.  I haven't tried it out, but it may require compartmentalizing the two library files.  That is something that should be done, but may want to be part of rethinking the API now that we have a better idea of how it is going to be used.  (We talked about this some time ago.)

pkra:
This fell of the radar.

a) it can't be merged anymore. 
b) it's aimed at master, not develop.

It would be great to restart the conversation though.

Issue #59, opened 2014-12-11
Allow change of svg-to-PNG converter [was: Why not use "svg2png" , pure node without JAVA depend]


everyx:
I hava some trouble in my vps about java environment configuration, and i find the node package "svg2png" and modify some code, i'll  submit a pull request about this for a review.

sorry for my poor english.

pkra:
See also https://github.com/mathjax/MathJax-node/issues/65#issuecomment-73074306

SamyPesse:
:+1:  [svgexport](https://www.npmjs.com/package/svgexport) works great

pkra:
@SamyPesse see [my earlier comment](https://github.com/mathjax/MathJax-node/issues/59#issuecomment-66744347) regarding our original requirements.

physikerwelt:
I agree that both solutions java and phantomjs are not ideal.
I think one of svgexports advantages is that no file needs to be written on "disk".
A node only solution would be awesome, but I guess it's a very challanging tasks, which can not be done a few after work sessions.
(ping @gwicke)

pkra:
Some recent observations

* Over on the Wikimedia fork, there's a PR building on rsvg, https://github.com/wikimedia/MathJax-node/pull/16. Biggest immediate question from my end was: is that a bad dependency?
* But I just noticed that [jsdom now handles canvas](https://github.com/tmpvar/jsdom#canvas) via https://www.npmjs.com/package/canvas. That seem to me to be the most natural approach ("universal JS" etc) despite having similarly bad installation requirements on node.

pkra:
Note: I haven't actually tested if the canvas solution for jsdom would be sufficient.

gwicke:
Some quick notes from my end:

- We use (a patched version of) rsvg for SVG thumbnailing in production. It is working okay, but lacks support for some features like HTML-in-SVG. There is some RSVG-related discussion on https://phabricator.wikimedia.org/T40010.
- Latest phantomjs (2.0) is based on relatively recent Blink, so likely has similar SVG support as current Chrome. However, binding support for node is mostly still using phantomjs 1.9. We don't currently use phantomjs in production, but might reconsider it for tasks like HTML-to-PDF in the future.
- I can't vouch for the quality or performance of the jsdom canvas solution. The dependencies are manageable (we have a few services using it).

d00rman:
[...]

PR #81, opened 2015-03-30
Specifiy batik version

physikerwelt:
The latest version batik 1.8 did not work for me.

### DIFF
--- a/batik/README.md
+++ b/batik/README.md
@@ -1,5 +1,5 @@
 If you intend to use MathJax-node's ability to create PNG images, you
-should install [batik](http://xmlgraphics.apache.org/batik/download.html)
-in this directory.  Just download it and unpack it here.  You need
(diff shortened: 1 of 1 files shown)


physikerwelt:
I think this patch should land in 0.3.
Batik 1.8 can be tested in the future.
See the error I got for 1,8
(https://gist.github.com/physikerwelt/57b3884fb89b475d1994)

pkra:
Please make PRs against `develop` not `master.

pkra:
FWIW, it seems like the problem is that Batik changed their naming scheme (to `-1.8.jar`) and renaming the files breaks things. If you change mj-single/page to match the file names, things work. Yet another reason for #59...

dpvc:
You can make a symbolic link named `batik-rasterizer.jar` to the `batik-rasterizer-1.8.jar` file, and MathJax-node will find and use that just fine.  Not sure what the Windows version of that is, but this works on Mac OS X, and presumably, Linux

PR #84, opened 2015-03-30
Specifiy batik version

physikerwelt:
The latest version batik 1.8 did not work for me.

### DIFF
--- a/batik/README.md
+++ b/batik/README.md
@@ -1,5 +1,5 @@
 If you intend to use MathJax-node's ability to create PNG images, you
-should install [batik](http://xmlgraphics.apache.org/batik/download.html)
-in this directory.  Just download it and unpack it here.  You need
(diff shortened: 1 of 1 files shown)


pkra:
See my comments on #81.

dpvc:
I think mentioning the symbolic link option might be a better solution to the 1.8 problem.

physikerwelt:
OK it works on linux... but windows does not support symlinks.
Feel free to regard this pull request as short bug report... so you probably don't want to merge it.

physikerwelt:
Even in the new version png generation is extremly slow
´´´
 time ./tex2png 'E=mc^2' > ~/test.png

real    0m3.421s
user    0m2.929s
sys     0m0.172s
´´´

pkra:
@physikerwelt do you happen to have any comparable data, e.g., using svg2png?

PR #90, opened 2015-04-24
Equations and ids

meghanlarson:
The main changes that have been made are:

* Add the equations option for mj-page.js. This allows users to receive a list of all equations that were converted in the document, including the source asciiMath, LaTex, or MathML.
* Error handling. Previously, if MathJax received an error, the API call would hang and never return. Now, you get a list of any MathJax errors that were encountered.
* IDs on element in the page. This creates an easy way to associate the equations with the elements in the converted document.


pkra:
hey @meghanlarson was this closed accidentally?

dpvc:
@pkra, Meghan had contacted me privately to ask how the pull process works (how quickly we respond, do we need use cases and sample files, etc), and I mentioned in my response that this should be from develop to develop, so she has close it and will submit another pull request later.

pkra:
Ah, thanks for the context.

PR #91, opened 2015-04-27
Equations and ids

meghanlarson:
The main changes that have been made are:

* Add the equations option for mj-page.js. This allows users to receive a list of all equations that were converted in the document, including the source asciiMath, LaTex, or MathML.
* IDs on element in the page. This creates an easy way to associate the equations with the elements in the converted document.


pkra:
@meghanlarson we'd need a CLA from you (http://www.mathjax.org/cla.html) and also, I'm assuming, a cCLA from Benetech, http://www.mathjax.org/ccla.txt.

(I'm also not sure why there's a commit from myself in there.)

dpvc:
@meghanlarson, thanks for making a new pull request, but there is still a problem.  I should have been more explicit in my message to you earlier.  The issue is not that your branch has to be named `develop`l it is rather that your branch *originate* from the MathJax-node `develop` branch not from the  MathJax-node `master branch.  The git-flow model only allows pull request from `develop` to `master`, not the other way around, and since your `develop` branch is just a branch off of your `master` branch, which is a branch off of MathJax-node's `master` branch, your pull request effectively pulls `master` into `develop`, which is the wrong direction.  I know that `develop` didn't exist when you started your work, but now that we are using the more formal git-flow model, we want to try to maintain it properly.  (@pkra, this is the reason your commit is there, because it was on our `master` branch that was later merged into the Benetech `master` branch.)

If you look at the [network diagram](https://github.com/mathjax/MathJax-node/network), you will see the black `master` branch and the blue `develop`.  Since the develop branch was created in December last year, all the changes have been to `develop`, and the pulls form loops that come from `develop` and to `develop`.  There is one pull to master (the last commit there) from `develop` to `master`.  That marked the latest release.  but everything else should be like the loops on the blue `develop` branch.

So what you normally would need to do is fetch the latest version from the MathJax-node repository, checkout the MathJax-node `develop` branch, creating a new `develop` branch in your local repository (you will need to delete or rename your current `develop` branch).  Then you would rebase your changes onto that branch.  I would take that opportunity to remove the commits that added and removed the `batik` library, so we don't have that as part of the history of MathJax-node.  That would take some cherry-picking or several rebase commands, but I would prefer not having those in the MathJax-node repository.

On the other hand, since you are only changing one file, it would probably be easier just to fetch the current MathJax-node repository, rename your `develop` branch and check out MathJax-node's `develop` branch, then replace the `lib/mj-page.js` file with your current one, commit that, and make a new pull request.

I am also going to ask you to change the code that skips processing when there are error messages, as there are reasons that there may be error messages but you still want to process the other math on the page.  For example, the use of characters missing from the MathJax fonts will cause error messages, but you still want the results processed.  So those `if-then` statements will need to be removed.

Once that is done, I'll be able to see your changes better and can review the pull request more carefully.  Sorry for asking you to change the pull request yet again!

meghanlarson:
Sure, I can do that. The add-remove of batik has been bugging me as well (I initially put it in because I knew we would always want to have batik installed, but it can just as easily be managed through our Vagrant file). 

Thanks for the help,
Meghan

meghanlarson:
Closing to create a pull request that follows the correct git-flow model.

meghanlarson:
Sorry for the hassle of getting this pull request correct. I think this one should be correct: https://github.com/mathjax/MathJax-node/pull/91

Thanks,
meghan

PR #112, opened 2015-08-12
update jsdom for iojs / nodejs v4 compatibility

### LINKED ISSUE #110: Making iojs the default
We should consider switching to iojs as default. 

Nodejs and iojs are due to merge soon with a first LTS release in October. Our primary dependency (jsdom) is now iojs only (with numerous benefits and fewer dependencies, especially node-gyp).

pkra:
Updates jsdom to enable upcoming node+iojs merge (v4). 

:warning:  This drops compatibility for older nodejs versions.

Closes #110.

### DIFF
--- a/lib/mj-page.js
+++ b/lib/mj-page.js
@@ -116,7 +116,7 @@ var STYLES;           // filled in when SVG is loaded
 function GetWindow() {
   document = jsdom();
   html = document.firstChild;
-  window = document.parentWindow;
+  window = document.defaultView;
   window.console = console;
   window.onerror = function (err,url,line) {AddError("Error: "+err)}
   content = document.body.appendChild(document.createElement("div"));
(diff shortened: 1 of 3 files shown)


pkra:
We should probably discuss this before merging.

dpvc:
Is this ready to go?  You have said we should discuss before merging.  Your pull requests for the tape tests includes this, so I wasn't sure if there was anything else to say about it first.

It would be possible to use

    window = document.defaultView || document.parentWindow;

(or in the other order), to make it compatible with the older jsdom version.  Is there some other concern about older nodejs versions?  I don't see anything else that would prevent that (other than jsdom's version number being `*` now).

pkra:
> Is this ready to go? 

I think so. But...

> It would be possible to use `window = document.defaultView || document.parentWindow;`

... I didn't think of that.

> Is there some other concern about older nodejs versions?

Not that I'm aware of. 

> Your pull requests for the tape tests includes this

I admit, mostly because that's what I was working on.

dpvc:
It looks like `document.defaultView` also works in jsdom 3.1.1, so I think we can leave it as is for now.  I am going to merge this, and then the tape one.

PR #116, opened 2015-09-08
adding random number to hard coded filename to allow concurrency

onomojo:
When you have more than 1 request hitting the server at the exact same time, they both try to access the same hard coded filename. This just adds some randomness to the filename so the concurrent requests don't try to use the same temp file.

### DIFF
--- a/lib/mj-page.js
+++ b/lib/mj-page.js
@@ -87,7 +87,7 @@ var MathJax;                       // filled in once MathJax is loaded
 var serverState = STATE.STOPPED;   // nothing loaded yet
 var timer;                         // used to reset MathJax if it runs too long
 
-var tmpfile = os.tmpdir() + "/mj-single-svg";  // file name prefix to use for temp files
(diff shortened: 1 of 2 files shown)


dpvc:
The usual approach to this would be to append the process PID rather than a random number, since that is guaranteed to be unique among running processes.  I would recommend changing the random number to `process.pid` instead.

It sounds like your server code is running Mathjax-node in several child processes.  It would be inefficient to start up MathJax-node for each request to the server (as MathJax startup time is non-trivial).  That is why MathJax-node is designed to processes multiple requests, not just one.  It may be OK to start up several children to service requests as long as they are able to stay alive and handle multiple requests each.

pkra:
@onomojo please rebase to `develop` and make the PR to `develop` as well.

onomojo:
I opened a new PR on the develop branch. https://github.com/mathjax/MathJax-node/pull/117 Closing this one.

PR #117, opened 2015-09-11
adding process id to tmp file to allow for concurrency

onomojo:
When you have more than 1 request hitting the server at the exact same time, they both try to access the same hard coded filename. This just adds the process id to the filename so the concurrent requests don't try to use the same temp file.

### DIFF
--- a/lib/mj-page.js
+++ b/lib/mj-page.js
@@ -87,7 +87,7 @@ var MathJax;                       // filled in once MathJax is loaded
 var serverState = STATE.STOPPED;   // nothing loaded yet
 var timer;                         // used to reset MathJax if it runs too long
 
-var tmpfile = os.tmpdir() + "/mj-single-svg";  // file name prefix to use for temp files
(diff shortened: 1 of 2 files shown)


dpvc:
@pkra, do we need a CLA for this?  I'm ready to merge the changes.

pkra:
> @pkra, do we need a CLA for this? I'm ready to merge the changes.

CLA is not necessary for this.

Do we need a test?

dpvc:
> Do we need a test?

I'm not sure I would bother with one for this.  If you wanted to do one, it would require running two copies of MathJax-node asynchronously so that they both are writing SVG files at once.  The timing issues seem hard to manage to guarantee that a test really does what it is supposed to do.

dpvc:
=> Merged.

pkra:
>  If you wanted to do one,

perhaps @onomojo is willing to add one -- it's your PR :-)

PR #129, opened 2015-10-28
Update Readme

pkra:
Change package name to new form

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# MathJax-node [![Build Status](https://travis-ci.org/mathjax/MathJax-node.svg?branch=develop)](https://travis-ci.org/mathjax/MathJax-node)
+# mathjax-node [![Build Status](https://travis-ci.org/mathjax/MathJax-node.svg?branch=develop)](https://travis-ci.org/mathjax/MathJax-node)
 
 This repository contains files that provide APIs to call MathJax from 
 node.js programs.  There is an API for converting individual math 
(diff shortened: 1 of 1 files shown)


pkra:
Question: should we rename the repository?

dpvc:
> Question: should we rename the repository?

My gut feeling is "no".  I'm not sure of the ramifications of doing so, as people have it checked out now under the current name, and I assume that would break any checked out clones.

It seems to me that as long as the `npm` package name is lower case, the actual repository doesn't matter, right?

README.md
```diff
@@ -25,7 +25,7 @@ directory for more details.
 
 # Getting started
 
-MahJax-node provides two libraries, `lib/mj-single.js` and `lib/mj-page.js`. Below are two  very minimal examples -- be sure to check out the examples in `./bin/` for more advanced configurations.
+mathjax-node provides two libraries, `lib/mj-single.js` and `lib/mj-page.js`. Below are two  very minimal examples -- be sure to check out the examples in `./bin/` for more advanced configurations.
```
dpvc: One of the reasons I've always preferred mixed-case names is that if the product name starts with a capital letter, you can use it at the beginning of a sentence without it looking wrong. I always regretted that `jsMath` started with lower case letters, since I had to reword things to avoid using it at the beginning of a sentence.  Oh well.

pkra: >  Oh well.

+1

PR #132, opened 2015-12-10
Test corner case that broke the wmf fork

physikerwelt:
This example breaks the MathJax-node wmf fork
See the output on
https://travis-ci.org/physikerwelt/MathJax-node/builds/95994288
Math Processing Error: Maximum call stack size exceeded
Error: Maximum call stack size exceeded
Thereafter, MathJax-node does not do anything

### DIFF
--- a/test/suicide.js
+++ b/test/suicide.js
@@ -1,0 +1,17 @@
+var tape = require('tape');
+var mjAPI = require("..//lib/mj-single.js");
+
+tape('basic test: check MathJax core', function(t) {
+    t.plan(1);
+
+    var tex = "\\underbrace {a \\choose b}";
+    mjAPI.start();
+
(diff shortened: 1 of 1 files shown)


physikerwelt:
@dpvc can you have a look at his issue?

pkra:
The test isn't actually valid (there's no outcome) but I can reproduce the problem once it's fixed.

It seems to be an upstream problem, since it doesn't happen with MathJax v2.6-beta1. 

I'm guessing it's https://github.com/mathjax/MathJax/issues/1151. So it comes down to the fact that we didn't update mathjax-node to MathJax v2.5.2. Sorry about that!

I'm closing this as it's an upstream bug. I've opened #133 instead.

PR #137, opened 2015-12-29
Update Readme

### LINKED ISSUE #136: change README to show the code as people would use it?
Right now the README.md shows a use of `require` that references local files, rather than files that are part of the module: `require("./lib/mj-page.js");`. This implies someone has to manually copy the `mj-page.js` file to their local `lib` dir, whereas these files are already part of the module's payload and can simply be required the normal way. It would be better to change this to `require("MathJax-node/lib/mj-page.js");` since that is how people will have received the mathjax-node package.

Although related to that, it's worth renaming the package for the next release to all-lowercase, to prevent possible problems with case sensitive operating systems (and a trivial refactor for users, since virtually all code editors come with a "replace in all files" function)

pkra:
;Improve examples; fixes #136

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ directory for more details.
 
 mathjax-node provides two libraries, `./lib/mj-single.js` and `./lib/mj-page.js`. Below are two  very minimal examples -- be sure to check out the examples in `./bin/` for more advanced configurations.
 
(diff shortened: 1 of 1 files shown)


pkra:
Fixes #136

PR #154, opened 2016-02-04
Add more patches for padding and margin

dpvc:
This makes setting individual margin or padding sides also update the general margin or padding style.


pkra:
:+1:

dpvc:
CSSStyleDeclaration has merged my changes, so the patches probably aren't needed (but there are checks to see if they are, and the patches aren't applied if not needed).

PR #155, opened 2016-02-05
update .travis.yml

pkra:
* extend test coverage of nodejs; fixes #153
* add automatic publishing to npm; fixes #152

PR #165, opened 2016-02-10
Release v0.5.0-rc.1

pkra:


pkra:
Huh? We have a merge problem?

pkra:
It's the travis file; I don't see why but oh well.

dpvc:
The issue seems to be the pull request #128, which was to master rather than to develop, and was never merged into develop.  It changed the node version in travis tests to 4.1.  The current travis file has set that back to 4, so that line is different in the two branches and is the source of the conflict.  I merged 128 into develop, and tested it by hand first to be sure that develop would then merge cleanly, and it did when I do the merge by hand.  But it still says that it doesn't merge here.

I also tried changing that line back to 4.1, and seeing that that allowed the merge.  Again, it does by hand, but not here.  So I'm unsure what is going on with GitHub's PR that makes it different from standard merges.  We may have to do the merge by hand rather than here.

The only other thing I could think of is to use git rebase to remove PR #155, then merge with 128 again, then reapply #155.  That can be done, but it would require unprotecting develop temporarily since it will require a forced push.

I guess we have to be more careful about not making changes to master that aren't reflected in develop.

pkra:
Thanks for tracking this down and apologies for causing it in the first place.

Let's do it by hand then once #167 and #168 are merged.

pkra:
Merged manually as discussed.

PR #167, opened 2016-02-10
package.json: point "main" to mj-single

### LINKED ISSUE #162: package.json default
I have a question. 

[Currently](https://github.com/mathjax/MathJax-node/blob/master/package.json#L33), `mj-page` is the default. I was wondering if mj-single is not the more commonly used one (as long as we don't have a single solution). 

Any thoughts?

pkra:
Fixes #162

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   "scripts": {
     "test": "tape test/*.js"
   },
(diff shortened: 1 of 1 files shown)


zorkow:
LGTM.

PR #174, opened 2016-02-18
[WIP] Modularization of PNG generation

### LINKED ISSUE #59: Allow change of svg-to-PNG converter [was: Why not use "svg2png" , pure node without JAVA depend]

pkra:
* [ ] abstract the png generation 
  * [X] mj-single
  * [ ] mj-page 
* [X] move batik-specific code into a separate module
* [X] add librsvg and svg2png based modules.
* [ ] decide on new default
  * update package.json
* [ ] update Readme

This fixes #59.

### DIFF
--- a/lib/mj-batik.js
+++ b/lib/mj-batik.js
@@ -1,0 +1,28 @@
+var fs = require('fs');
+var os = require('os');
+var path = require('path');
+var execFile = require('child_process').execFile;
+
+exports.svg2png = function(svg, result, scale, width, height, callback, check) {
+  var tmpfile = os.tmpdir() + "/mj-svg" +  process.pid;  // file name prefix to use for temp files
+  var BatikRasterizerPath = path.resolve(__dirname,'..','batik/batik-rasterizer.jar');
+  var batikCommands = ['-jar', BatikRasterizerPath, '-dpi', scale*72, tmpfile + '.svg'];
+  var tmpSVG = tmpfile + ".svg",
+    tmpPNG = tmpfile + ".png";
+  fs.writeFile(tmpSVG, svg, function(err) {
+    if (check(err)) return;
+    execFile('java', batikCommands, function(err, stdout, stderr) {
+      if (check(err)) {
+        fs.unlinkSync(tmpSVG);
+        return
+      }
+      fs.readFile(tmpPNG, null, function(err, buffer) {
+        result.png = "data:image/png;base64," + (buffer || "").toString('base64');
+        fs.unlinkSync(tmpSVG);
+        fs.unlinkSync(tmpPNG);
+        check(err);
+        callback();
+      });
+    });
+  });
+}
--- a/lib/mj-librsvg.js
+++ b/lib/mj-librsvg.js
@@ -1,0 +1,26 @@
+var rsvg = require('librsvg').Rsvg;
+var Readable = require('stream').Readable;
+
+exports.svg2png = function(svg, result, scale, width, height, callback, check) {
+  var s = new Readable();
+  var svgRenderer = new rsvg();
+  s._read = function() {
+    s.push(svg);
+    s.push(null);
+  };
+  svgRenderer.on('error', function(err){
+    check(err);
+    return;
+  });
+  svgRenderer.on('finish', function() {
+    // console.log(width, height, scale, svg);
+    var buffer = svgRenderer.render({
+      format: 'png',
+      width: width * scale,
+      height: height * scale
+    }).data;
+    result.png = "data:image/png;base64," + (buffer || "").toString('base64');
+    callback();
+  });
+  s.pipe(svgRenderer);
+};
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -28,13 +28,11 @@
 
 var http = require('http');
 var fs = require('fs');
-var path = require('path');
 var url = require('url');
 var fmt = require('util').format;
 var jsdom = require('jsdom').jsdom;
-var execFile = require('child_process').execFile;
 var speech = require('speech-rule-engine');
-var os = require('os');
+var svg2png = require('./mj-batik.js').svg2png;
 
 require('./patch/jsdom.js').patch(jsdom);  //  Fix some bugs in jsdom
 
@@ -63,6 +61,7 @@ var defaults = {
   img: false,                     // return img tag for remote image?
   png: false,                     // return png image (as data: URL)?
   dpi: 144,                       // dpi for png image
+  scale: 2,                       // scale for png image
 
   speakText: false,               // add spoken annotations to svg output?
   speakRuleset: "mathspeak",      // set speech ruleset (default (chromevox rules), mathspeak)
@@ -85,12 +84,10 @@ var STATE = {
 // The MathJaxPath is normaized against file:/// so that Windows paths are correct
 //
 var MathJaxPath = url.resolve("file:///","file:"+require.resolve('mathjax/unpacked/MathJax'));
-var BatikRasterizerPath = path.resolve(__dirname,'..','batik/batik-rasterizer.jar');
 var MathJaxConfig;                   // configuration for when starting MathJax
 var MathJax;                         // filled in once MathJax is loaded
 var serverState = STATE.STOPPED;     // nothing loaded yet
 var timer;                           // used to reset MathJax if it runs too long
-var tmpfile = os.tmpdir() + "/mj-single-svg" +  process.pid;  // file name prefix to use for temp files
 
 var document, window, content, html; // the DOM elements
 
@@ -604,7 +601,11 @@ function GetSVG(result) {
   //  Add the requested data to the results
   //
   if (data.svg) result.svg = svgdata;
-  if (data.png) result.svgfile = svgfile;
+  if (data.png) {
+    result.svgfile = svgfile;
+    result.width = svg.getAttribute("width").replace("ex","")*jax.SVG.ex;
+    result.height = svg.getAttribute("height").replace("ex","")*jax.SVG.ex;
+  }
   if (data.img) {
     if (data.svg) result.svg = svgfile;
     result.img = [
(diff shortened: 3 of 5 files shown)


pkra:
@dpvc @zorkow the main changes are ready for review. I marked it WIP for the organizational details.

pkra:
FYI, the tests fail because Batik won't be installed on Travis at the moment. They pass locally for me.

pkra:
To clarify: the PR currently only modifies mj-single. Once that part is reviewed, I'll reproduce it for mj-page.

pkra:
While the test that's in this PR is universal (in that it succeeds with all three converter modules), I'm wondering if this approach is not actually taking things far enough. 

Should the PNG conversion be ripped out completely instead? Just pass the SVG to your own lib and done. It's a bit radical, I admit. But for mj-single it seems sensible; PNG generation is really not mathjax-y.

lib/mj-librsvg.js
```diff
@@ -0,0 +1,26 @@
+var rsvg = require('librsvg').Rsvg;
+var Readable = require('stream').Readable;
+
+exports.svg2png = function(svg, result, scale, width, height, callback, check) {
+  var s = new Readable();
+  var svgRenderer = new rsvg();
```
d00rman: You can avoid the stream altogether here since you already have the data to render in `svg`, so simply:

```js
var svgRender = new rsvg(svg);
```

d00rman: Note that this still allows you to keep the block on L11 (error checking), because the `rsvg` object is a stream itself.

d00rman:
> Should the PNG conversion be ripped out completely instead? Just pass the SVG to your own lib and done. It's a bit radical, I admit. But for mj-single it seems sensible; PNG generation is really not mathjax-y.

I'm inclined to think the same. Conversion in itself is not (or should not, IMO) be MathJax' problem. But, a pragmatic solution might be to allow users to specify which converter to use in the init phase. Ideally, we'd want the conversion engine or lib to be a *peer dependency* of MathJax, i.e. it should not be required by MathJax itself, because, as we see from this PR, that entails a strong dependency on either:
- a binary module / package; or
- a large Java-based solution

If there were a really slow, but light dependency that could be used as the sane default, that would be awesome, but I have a feeling that here explaining to users how to achieve the conversion is the right way.

In that context clearly defining an interface for the conversion (which I think has been achieved implicitly in this PR) is a good practice, because that means additional converters can be only an `npm install` away, as it makes it rather easy to package different converters separately: one can easily imagine `lib/mj-batik` and `lib/mj-librsvg` being separate npm packages, and then the user just supplies them to MathJax during the initialisation phase.

lib/mj-librsvg.js
pkra: Thanks! Since I'm not too experienced with streams: does L11 catch only errors from the streaming process (which is nice if the string was corrupted somehow) or does it also catch errors that rsvg itself has produced?

pkra:
@d00rman thanks for the comments!

>  Ideally, we'd want the conversion engine or lib to be a peer dependency of MathJax

I've never used peer dependencies. How would that work?

> one can easily imagine lib/mj-batik and lib/mj-librsvg being separate npm packages, and then the user just supplies them to MathJax during the initialisation phase.

I think I agree but I'm not sure. FWIW, I had been thinking a separate module `mj-svg2png` could look something like

```
var mj = require('./lib/mj-single');
var svg2png = require('svg2png');

exports.mj2png = function(options, callback) {
  mj.start();
  mj.typeset(options, function(data) {
    var sourceBuffer = new Buffer(data.svg, "utf-8");
    var returnBuffer = svg2png.sync(sourceBuffer, {
      width: data.width,
      height: data.height
    });
    data.png = "data:image/png;base64," + returnBuffer.toString('base64');
    callback(data);
  });
};
```

(note: just a mockup).

lib/mj-batik.js
```diff
@@ -0,0 +1,28 @@
+var fs = require('fs');
+var os = require('os');
+var path = require('path');
+var execFile = require('child_process').execFile;
+
+exports.svg2png = function(svg, result, scale, width, height, callback, check) {
+  var tmpfile = os.tmpdir() + "/mj-svg" +  process.pid;  // file name prefix to use for temp files
+  var BatikRasterizerPath = path.resolve(__dirname,'..','batik/batik-rasterizer.jar');
+  var batikCommands = ['-jar', BatikRasterizerPath, '-dpi', scale*72, tmpfile + '.svg'];
```
dpvc: You've lost the use of the `dpi` setting that is part of the current API by changing it to `scale*72` here, so that is a breaking change.  (Also, the default dpi passed by the command-line tools is based on the ex size, with the default ex of 6 producing a dpi of 96, so this will produce different sized images from the current version.  I worked a bit to get the `dpi` values to do the right thing for the given ex size.)

lib/mj-single.js
```diff
@@ -63,6 +61,7 @@ var defaults = {
   img: false,                     // return img tag for remote image?
   png: false,                     // return png image (as data: URL)?
   dpi: 144,                       // dpi for png image
+  scale: 2,                       // scale for png image
```
dpvc: Why is this 2 by default?  Wouldn't a normal default scaling be 1?  And how does this relate to the dpi value?  Shouldn't scaling be determined by the dpi and ex values?

lib/mj-batik.js
dpvc: If you want to use a width and height rather than dpi, batik can do that (according to the man page), so it would be possible to use those instead (and change or remove the command line dpi API, or base the scaling off the dpi, e.g., make scale = dpi / 96).  In that case, however, I'd only pass width and height and have the scale multiplied in before `svg2png` is called so that the `width` and `height` here are the true sizes that are being requested.

lib/mj-single.js
```diff
@@ -622,23 +623,20 @@ function GetSVG(result) {
...
-    var batikCommands = ['-jar', BatikRasterizerPath, '-dpi', data.dpi, tmpfile + '.svg'];
+    var pngScale = data.scale;
+    var svgFile = result.svgfile; delete result.svgfile;
+    var width = result.width; delete result.width;
+    var height = result.height; delete result.height;
+    var check = function(err) {
+      if (err) {
+        AddError(err.message);
+        synch();
+        return true
```
dpvc: should have a trailing semi-colon here.

lib/mj-single.js
```diff
@@ -622,23 +623,20 @@ function GetSVG(result) {
...
 //
 function GetPNG(result) {
-  var svgfile = result.svgfile; delete result.svgfile;
   if (data.png) {
-    var batikCommands = ['-jar', BatikRasterizerPath, '-dpi', data.dpi, tmpfile + '.svg'];
+    var pngScale = data.scale;
+    var svgFile = result.svgfile; delete result.svgfile;
+    var width = result.width; delete result.width;
+    var height = result.height; delete result.height;
+    var check = function(err) {
```
dpvc: I prefer to see the `check` function *after* the `synch` function has been defined, but it is not actually an error.

lib/mj-batik.js
```diff
@@ -0,0 +1,28 @@
...
+  var tmpSVG = tmpfile + ".svg",
+    tmpPNG = tmpfile + ".png";
+  fs.writeFile(tmpSVG, svg, function(err) {
+    if (check(err)) return;
+    execFile('java', batikCommands, function(err, stdout, stderr) {
+      if (check(err)) {
+        fs.unlinkSync(tmpSVG);
+        return
```
dpvc: should have a trailing semi-colon here when you separate the commands out to separate lines.

lib/mj-svg2png.js
```diff
@@ -0,0 +1,12 @@
+var fs = require('fs');
+var svgpng = require('svg2png');
+
+exports.svg2png = function(svg, result, scale, width, height, callback) {
+  var sourceBuffer = new Buffer(svg, "utf-8");
+  var returnBuffer = svgpng.sync(sourceBuffer, {
```
dpvc: The `sync` version here will cause the mathjax-node process to block during the conversion and not be able to receive additional requests during that time.  We have avoided the use of synchronous calls (except in cases where they should complete immediately, like removing temporary files) so that if mathjax-node is used as part of something like a web service, that service can continue to respond to requests (by queuing mathjax-node typeset calls) even while mathjax-node is handling earlier requests.  This will prevent that from happening during png creation.

test/png.js
```diff
@@ -0,0 +1,18 @@
...
+    t.plan(2);
+
+    var tex = '\\sin(x)';
+    mjAPI.start();
+
+    mjAPI.typeset({
+        math: tex,
+        format: "inline-TeX",
+        png: true,
+        mathoidStyle: true,
```
dpvc: I don't recognize this parameter.  What does it mean?

pkra:
@dpvc thanks for the review. But before I fix this, perhaps we can first discuss the larger question of moving the entire PNG workflow out of mathjax-node itself.

dpvc:
I'm OK with removing the PNG stuff (it was a pain to implement), though I suspect you will get some push-back from people currently using it.  It might be nice to keep some `*2png` command-line tools (but with the svg-to-png conversion being done through an external tool in the command-line program rather than in mj-single itself).

test/png.js
physikerwelt: It's not used in the test. So it can be removed. mathoidStyle is a modified version of img which only returns the style

lib/mj-librsvg.js
d00rman: According to the lib's code, both would trigger L11.

pkra: Thanks, @d00rman !

d00rman:
> I've never used peer dependencies. How would that work?

NPM supports [`peerDependencies`](https://nodejs.org/en/blog/npm/peer-dependencies/) where one can specify the modules that should be included by the application. A big note here is that the `peerDependencies` sections will disappear with npm 3.x.

The general idea is that an external package provides the `svg2png` functionality, which is then related to MathJax-node when `require()`-ing it (or setting a global variable that would be picked up by it).

> I think I agree but I'm not sure. FWIW, I had been thinking a separate module mj-svg2png could look something like
> 
> var mj = require('./lib/mj-single');
> var svg2png = require('svg2png');
> 
> exports.mj2png = function(options, callback) {
>   mj.start();
>   mj.typeset(options, function(data) {
>     var sourceBuffer = new Buffer(data.svg, "utf-8");
>     var returnBuffer = svg2png.sync(sourceBuffer, {
>       width: data.width,
>       height: data.height
>     });
>     data.png = "data:image/png;base64," + returnBuffer.toString('base64');
>     callback(data);
>   });
> };
> (note: just a mockup).

Yup, something like that :+1:

pkra:
>  A big note here is that the peerDependencies sections will disappear with npm 3.x.

Ah, right. I'm too used to 3.x already, so I got confused when I read up on peerDepdencies. Makes sense now, thank you!

pkra:
Closing in favor of #205

Issue #179, opened 2016-02-20
Cannot find module 'cssstyle/lib/parsers.js'


a3nm:
Hi,

I tried to install mathjax-node on a fresh Debian testing install as follows:

```
# nodejs-legacy required for node to be available as "node" rather than "nodejs"
apt-get install npm nodejs-legacy
npm install mathjax-node
```

I then tried to run:

```
node_modules/mathjax-node/bin/am2html
```

It results in the following:

```
module.js:328
    throw err;
    ^

Error: Cannot find module 'cssstyle/lib/parsers.js'
    at Function.Module._resolveFilename (module.js:326:15)
    at Function.Module._load (module.js:277:25)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at /node_modules/mathjax-node/lib/patch/jsdom.js:51:17
    at Object.<anonymous> (/node_modules/mathjax-node/lib/patch/jsdom.js:94:3)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
```

It seems that this problem can be fixed by issuing:

```
npm install cssstyle
```

Is this because the dependency of "mathjax-node" on "cssstyle" is not declared?

dpvc:
Can you say what versions of `node` and `npm` you are using?  (Use `node --version` and `npm --version`.)

> Is this because the dependency of "mathjax-node" on "cssstyle" is not declared?

No, `cssstyle` is not a dependency of `mathjax-node`, it is a dependency of `jsdom`, which is a dependency of `mathjax-node`.  But it has some bugs, and `mathjax-node` tries to patch those.  There was a change in the directory structure used by `node` between versions 4 and 5, and so `mathjax-node` has some logic to try to locate the library in both possible hierarchies.  Apparently, this isn't working properly.  Can you do `find . -name "cssstyle"` within the `mathjax-node` directory and tell me where the `cssstyle` library has been located in your case?  (Since you installed it explicitly yourself, there may be two copies.)

dpvc:
I believe that I have located the issue.  It is the same one described in PR #172.  The `issue172` branch includes a patch.  You can make the corresponding changes to your copy of `mathjax-node/lib/patch/jsdom.js` and that should resolve the issue for you.

a3nm:
Thanks for your answer! Here are the versions (they are the ones packaged by Debian testing):

```
[...]

PR #172, opened 2016-02-17
Rolling jsdom back to 6.*

sjlu:
It look like in the most recent version 0.5.0, you guys are trying to patch jsdom. In their 7.* version, they no longer use cssstyle which causes errors when trying to load the MathJax module.

**The Error**
```
web-0 Error: Cannot find module 'cssstyle/lib/parsers.js'
web-0     at Function.Module._resolveFilename (module.js:326:15)
web-0     at Function.Module._load (module.js:277:25)
web-0     at Function._load (/usr/local/Cellar/nvm/0.30.1/versions/node/v4.2.6/lib/node_modules/pm2/node_modules/pmx/lib/transaction.js:62:21)
web-0     at Module.require (module.js:354:17)
web-0     at require (internal/module.js:12:17)
web-0     at /Users/sjlu/Code/testable/node_modules/mathjax-node/lib/patch/jsdom.js:51:17
web-0     at Object.<anonymous> (/Users/sjlu/Code/testable/node_modules/mathjax-node/lib/patch/jsdom.js:94:3)
```

**Installed modules for MathJax at 0.5.0**
```
speech-rule-engine@0.2.8 node_modules/speech-rule-engine
├── xpath@0.0.21
├── xmldom@0.1.22
└── commander@2.9.0 (graceful-readlink@1.0.1)

yargs@3.32.0 node_modules/yargs
├── camelcase@2.1.0
├── window-size@0.1.4
├── y18n@3.2.0
├── os-locale@1.4.0 (lcid@1.0.0)
├── decamelize@1.1.2 (escape-string-regexp@1.0.4)
├── string-width@1.0.1 (is-fullwidth-code-point@1.0.0, strip-ansi@3.0.0, code-point-at@1.0.0)
└── cliui@3.1.0 (wrap-ansi@1.0.0, strip-ansi@3.0.0)

tape@4.4.0 node_modules/tape
├── inherits@2.0.1
├── defined@1.0.0
├── resumer@0.0.0
├── has@1.0.1
├── function-bind@1.0.2
├── deep-equal@1.0.1
├── through@2.3.8
├── minimist@1.2.0
├── object-inspect@1.0.2
├── resolve@1.1.7
├── glob@5.0.15 (path-is-absolute@1.0.0, once@1.3.3, inflight@1.0.4, minimatch@3.0.0)
└── string.prototype.trim@1.1.2 (define-properties@1.1.2, es-abstract@1.5.0)

jsdom@7.2.2 node_modules/jsdom
[...]

PR #182, opened 2016-02-22
Make test for node hierarchy more robust.

dpvc:
The current test was dependent on the current directory.  My tests were done from the mathjax-node directory itself, but when it is installed with `npm install mathjax-node`, rather than getting a copy of mathjax-node from github and running `npm install` in it, you get a `node_modules` directory, and so the current directory is not the same as what I was using, and the test for node version 4 failed, and it tried to use the node-5 hierarchy.  This leads to the error reported in issues #172 and #179.

This patch uses a test that doesn't rely on the current directory.

### DIFF
--- a/lib/patch/jsdom.js
+++ b/lib/patch/jsdom.js
@@ -3,12 +3,8 @@
 //  between node 4 and node 5, so check which one we have.
 //
 
-var fs = require('fs');
-
-var PARSERS = 'jsdom/node_modules/cssstyle/lib/parsers.js';  // node 4 hierarchy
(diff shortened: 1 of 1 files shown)


pkra:
LGTM.

dpvc:
@pkra, I'm happy to merge.  Does that mean a new release of mathjax-node?

pkra:
>  Does that mean a new release of mathjax-node?

Sure. I had set up a milestone for "what's next" but this seems urgent, right?

dpvc:
It would be a good idea to get this out, I think.

pkra:
> It would be a good idea to get this out, I think.

:+1: from me. I will finish the PR to master and make a release.

Issue #181, opened 2016-02-20
page2html adds "<style id="MathJax_CHTML_styles">undefined</style>"


a3nm:
The following is added at the end of the `<head>` tag when passing an HTML document through page2html, I am not sure why:

```
<style id="MathJax_CHTML_styles">undefined</style>
```

What is problematic is that this gets added by page2html even if already present, so if page2html is applied again to its own output, the output grows, and becomes invalid (two elements with the same id).

pkra:
I cannot reproduce this problem. You might want to share a minimal example exhibiting the problem.

a3nm:
I can reproduce it already with the following minimal document, piped through `page2html`:

```
<!DOCTYPE html>
<html><head>
  <title>blah</title>
</head>
<body>
</body></html>
```

I get the following output:

```
<!DOCTYPE html>
<html><head>
  <title>blah</title>
<style id="MathJax_CHTML_styles">undefined</style></head>
<body>

</body></html>
```

Do you get something different?

dpvc:
This happens only with an HTML file that doesn't contain any math.  If it has math, then the `undefined` will end up being the styles that are needed for the MathJax output.  It is easy to patch this not to add it when there is no math.

On the other hand, you point out correctly that if you process the output of page2html a second time, you will get a second copy of the style element.  I guess I can have page2html remove the old one before adding it in.

a3nm:
OK, thanks for the explanation! Yes, I think it would be cleaner to not add this when there is no math, and to ensure otherwise that page2html does not add a second style element with the same ID as an existing one.

dpvc:
The `issue181` branch includes the two fixes discussed above.

a3nm:
Thanks a lot! I have applied the patch to my installed version, and I have checked that the useless `<style>` element does not get added to my minimal document, and that applying `page2html` on the output of a run of `page2html` did not add more `<style>` (indeed it results only in changes in whitespace).

So this looks good to me, my only question would then be whether those fixes shouldn't also be done on all bin/page* utilities (and maybe others)? I haven't tried to use them so I don't know, but I don't see why this fix should be specific to `page2html`.

Thanks again!

dpvc:
[...]

PR #184, opened 2016-02-22
Don't include a style in the output when there isn't any math on the page

dpvc:
Don't include a style in the output when there isn't any math on the page, and remove an old copy of the styles if there was one already, and add some comments about what is going on.

Resolves issue #181

### DIFF
--- a/bin/page2html
+++ b/bin/page2html
@@ -110,7 +110,7 @@ function processHTML(html,callback) {
     //
     //  Replace the body with the one containing the math.
     //
-    document.body.innerHTML = result.html;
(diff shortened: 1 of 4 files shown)


pkra:
LGTM.

pkra:
Maybe add a test?

dpvc:
In fact, I added *two* tests, and in thinking about the tests, realized that my original solution was problematic, so I fixed that, too.

pkra:
> In fact, I added two tests, and in thinking about the tests, realized that my original solution was problematic, so I fixed that, too.

:100:

PR #188, opened 2016-02-23
Adds semantic tree output mj-single and a demonstrator executable.

zorkow:
Exposes the semantic tree via MathJax-node.
I included the stree computations into the general pipeline. I still feel this is overkill if one wants to only translate TeX into a MML/Stree.
I admit that I have no clue what ConfigureMathJax and StartQueue really do and I would strongly argue for their refactoring into smaller reusable/composable methods.

### DIFF
--- a/bin/tex2stree
+++ b/bin/tex2stree
@@ -2,13 +2,13 @@
 
 /*************************************************************************
  *
- *  tex2mml
+ *  tex2stree
  *
  *  Uses MathJax to convert a TeX or LaTeX string to a MathML string.
  *
  * ----------------------------------------------------------------------
  *
- *  Copyright (c) 2014 The MathJax Consortium
+ *  Copyright (c) 2016 The MathJax Consortium
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
--- a/bin/tex2stree
+++ b/bin/tex2stree
@@ -1,0 +1,56 @@
+#! /usr/bin/env node
+
+/*************************************************************************
+ *
+ *  tex2mml
+ *
+ *  Uses MathJax to convert a TeX or LaTeX string to a MathML string.
+ *
+ * ----------------------------------------------------------------------
+ *
+ *  Copyright (c) 2014 The MathJax Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+var mjAPI = require("../lib/mj-single.js");
+
+var argv = require("yargs")
+  .demand(1).strict()
+  .usage("Usage: tex2stree [options] 'math' > file.(json|xml)",{
+    json: {
+      boolean: true,
+      describe: "return semantic tree in Json"
+    },
+    xml: {
+      boolean: true,
+      describe: "return semantic tree in XML"
+    }
+  })
+  .argv;
+
+mjAPI.config({});
+mjAPI.start();
+
+mjAPI.typeset({
+  math: argv._[0],
+  format: "TeX",
+  semantic: true
+}, function (data) {
+  if (!data.errors) {
+    if (argv.xml) {
+      console.log(data.streeXml);}
+    if (argv.json) {
+      console.log(JSON.stringify(data.streeJson));}
+    }
+});
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -525,7 +525,7 @@ function AddError(message,nopush) {
 //  into account)
 //
 function GetMML(result) {
-  if (!data.mml && !data.speakText) return;
+  if (!data.mml && !data.speakText && !data.semantic) return;
   var jax = MathJax.Hub.getAllJax()[0];
   try {result.mml = jax.root.toMathML('',jax)} catch(err) {
     if (!err.restart) {throw err;} // an actual error
@@ -556,10 +556,6 @@ function GetSpeech(result) {
 //
 function GetSemantic(result) {
   if (!data.semantic) return;
-  if (!result.mml) {
-    data.mml = true ;
-    GetMML(result);
-  }
   result.streeJson = speech.toJson(result.mml);
   result.streeXml = speech.toSemantic(result.mml);
 }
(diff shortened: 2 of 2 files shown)


bin/tex2stree
```diff
@@ -0,0 +1,56 @@
+#! /usr/bin/env node
+
+/*************************************************************************
+ *
+ *  tex2mml
```
dpvc: You should probably change this to tex2stree, and adjust the comment below that describes what it is.

bin/tex2stree
```diff
@@ -0,0 +1,56 @@
...
+ *
+ *  Uses MathJax to convert a TeX or LaTeX string to a MathML string.
+ *
+ * ----------------------------------------------------------------------
+ *
+ *  Copyright (c) 2014 The MathJax Consortium
```
dpvc: I guess this should be 2016 for this one.

lib/mj-single.js
```diff
@@ -549,6 +551,20 @@ function GetSpeech(result) {
...
 
 //
+//  Creates the semantic tree for the current element and attaches it as JSON
+//  and XML.
+//
+function GetSemantic(result) {
+  if (!data.semantic) return;
+  if (!result.mml) {
+    data.mml = true ;
+    GetMML(result);
```
dpvc: Unfortunately, `GetMML` can't be called this way, since it is asynchronous, and could return before the result is available.  That is why `GetMML` and the other such functions are in the `HUB.Queue()` since the queue will wait until they complete before doing the next item in the queue.

So you probably should change [line 528](https://github.com/mathjax/MathJax-node/blob/stree_output/lib/mj-single.js#L528) in `GetMML` to not return if `data.semantic` is set.

Then I would move your `GetSemantic` call to before the `GetSpeech` call, because `GetSpeech` removes the `result.mml` value at [line 549](https://github.com/mathjax/MathJax-node/blob/stree_output/lib/mj-single.js#L549) if it is not to be returned.  (That should probably be done in a separate function or perhaps in `ReturnResult`.)

dpvc:
> I would strongly argue for their refactoring into smaller reusable/composable methods.

I'm not sure what you have in mind in terms of reusability and composability, here.  The `ConfigureMathJax` function is mostly just setting the MathJax configuration.  The biggest part of this is modifying the MathJax SVG and CHTML output to be able to be used with jsdom (so that they don't try to make measurements to find the `em` and `ex` size, or the size of unknown characters, and stuff like that).  I'm not sure how these could be reused.  It would be possible to break some of it out into separate functions that get called from the configuration in ConfigureMathJax, but I'm not sure how that helps.  There is a bit that converts the arrays in the user configuration into the jsdom window arrays, and I suppose that could be reusable, but other than that, I don't see what is.

As for `StartQueue`, this clears the values used in MathJax from any previous run (since the same MathJax API variable can be used to call `Typeset()` more than once), and then sets up the MathJax queue to do the typesetting, and handle the asynchronous actions that are needed to complete the processing of the equation that has been passed to the `Typeset()` call.  Again, I'm not sure what of this needs to be reused or composed with others.  The timing issues are delicate (as I point out in the code comments above), so the use of the queue is pretty important.  The individual items in the queue can't really be called separately.  As for the code before setting up the queue, I suppose that some parts could be packaged into separate functions to make `StartQueue()` shorter, but I don't see them being used in other ways.  The use of `StartQueue()` must be synchronized with the other MathJax processing as well, since you can't be resetting MathJax while it is processing some other mathematics.

So if you could be more specific about what use you have in mind, I can try to break things up with that in mind.

Of course, for v3.0, MathJax-node will have to be completely rewritten, so I'm not sure how much effort should be put into it at the moment.

lib/mj-single.js
zorkow: Since there is not much more cleaning up to do (as far as I can see), I put the delete result.mml into the ReturnResult function now.

zorkow:
> configuration in ConfigureMathJax, but I'm not sure how that helps. There is a bit that converts the 
> arrays in the user configuration into the jsdom window arrays, and I suppose that could be reusable, 
> but other than that, I don't see what is.

I am not sure whether there is something reusable. I kind of hoped there might be.
So here was my use case: Use MathJax-node to translate TeX into the semantic tree. I.e., TeX comes in (or possibly AM) and semantic tree comes out. 
I figured that all I needed was the pipeline of methods: 
 ```Typeset, TypesetDone, GetMML, GetSemantic, ReturnResult```
So no need to configure renderers, PNG, SVG etc. Unfortunately, when I tried to put together some minimal functionality for that, I failed, mainly because I really could not understand what was going on in the configuration function.

Regardless whether or not anything is reusable, I would argue a method that is 476 lines (including whitespace and comments) therefore roughly equivalent to 8 pages in my emacs needs to be broken up.

> Of course, for v3.0, MathJax-node will have to be completely rewritten, so I'm not sure how much 
> effort should be put into it at the moment.

You are right. I believe for now we should not touch it anymore, not worth the time. But I will make sure not to let too large units get to code reviews for 3.0.

zorkow:
Damn I just screwed up and merged master instead of develop.....

zorkow:
That rebases things to develop. Should be fine now. @dpvc ptal.

dpvc:
> I figured that all I needed was the pipeline of methods: 
`Typeset, TypesetDone, GetMML, GetSemantic, ReturnResult`
So no need to configure renderers, PNG, SVG etc.

Unfortunately, that is not the case.  The `Typeset` call requires both input and output jax to be set up, so you need at least one of the SVG or CHTML output rendered to be configured.  The ConfigureMathJax function also handles hooking into the error system to trap errors properly in mathjax-node, disabling the on-screen message updates, setting appropriate parameters for delays and such, loading the extensions requested by the caller, and adding any user configuration.  All of that needs to be done, and can't be left out.

Remember that `mj-single` and `mj-page` are designed to process multiple calls (e.g., as the back-end for a web service), and so may process some calls that don't need all the functions, and others that do.  The `ConfigureMathJax()` is only done once, not once for every expression processed.  While the command-line tools only call mathjax-node once, that is not the only use case.

> I would argue a method that is 476 lines ... needs to be broken up.

Agreed.  It has expanded over time as features were added (e.g., adding CommonHTML output nearly doubled it, since much of the function is devoted to patching the output renderers to avoid measuring DOM elements).

> I believe for now we should not touch it anymore, not worth the time.

OK.

>  will make sure not to let too large units get to code reviews for 3.0.

Very good.

zorkow:
A related question: do we have a pretty printer for XML expressions in
MathJax?
I was thinking about adding an option to the binary for pretty XML and json.
Volker

pkra:
> do we have a pretty printer for XML expressions in MathJax?

There's https://github.com/mathjax/MathJax/blob/master/unpacked/extensions/MathMenu.js#L1191-L1224 but it is probably not what you have in mind.

dpvc:
> That rebases things to develop. 

Actually it doesn't.  You still have the pull from master, which is not allowed in the git-flow model.

stree_merge

You need to use `git rebase` to remove that merge.  Something like

    git checkout stree_output
    git reset --hard HEAD~1
    git rebase --onto a0afa57 945e565 stree_output
    git push --force stree_output

The `git reset` removes the merge of the develop branch (since that was the last action on that branch).  The `git rebase` removes the merge of master (keeping the one commit that was made after that).  The `git push` should update `stree_output` to the new version.  I'm not sure how it will show up in this pull request.  You might need to make a new one, but I'm hoping not.

This should leave a clean branch for the pull request.

dpvc:
> do we have a pretty printer for XML expressions in MathJax?

No.  There is the `toMathML()` function that does that for the internal MathML format, but not a function in general, no.

Issue #191, opened 2016-03-03
[discussion] major changes


pkra:
The team would like to propose several major changes to mathjax-node.

* move `mj-page` out of mathjax-node
  * maintain an equivalent module (and document others, e.g., XML, markdown)
  * this will require new implementations of MathJax's pre-processors and connects to our work towards MathJax v3.0.
* move PNG generation out of mathjax-node
  * cf. the discussion in #174
  * maintain equivalent module (and document others)
* move SRE integration out of mathjax-node
  * cf. the discussion in #121
  * maintain an equivalent module.

If you have any comments on these proposals, please comment!

pkra:
Closed in favor of #205, #206, #207.

PR #192, opened 2016-03-03
Adds pretty printing and minimisation option to semantic tree output.

zorkow:


pkra:
:+1: for a PR against `develop` :wink:

zorkow:
Did I do it wrong again?

bin/tex2stree
```diff
@@ -49,8 +53,9 @@ mjAPI.typeset({
 }, function (data) {
   if (!data.errors) {
     if (argv.xml) {
-      console.log(data.streeXml);}
+      console.log(argv.min ? data.streeXml : data.pprintXml(data.streeXml));}
```
dpvc: It seems a bit odd to me to pass the function back in the data object.  Why not pass the `min` parameter in the typeset call and have `GetSemantic` send back the form that the user wants?  (I would recommend making the parameter to `mj-single` a bit more descriptive, like `minSTree` or something, since `min` is a bit generic (though the command-line argument of `--min` is fine).

dpvc:
> Did I do it wrong again?

No, you're good.  I think he was complimenting you (while giving you a bit of a dig at the same time).

zorkow:
Sure, just make fun of the guy coming home from the pub!

pkra:
> Sure, just make fun of the guy 

:bow: my humble apology.

zorkow:
I am always reluctant to add more parameters the main module has to be aware of in advance just to achieve some post processing.
In many ways the sre api should also be available outside mj-single. But this should be possible once we have figured out issue #191.

pkra:
In the mean time, could this benefit from a test?

dpvc:
> In many ways the sre api should also be available outside mj-single. And just passing it through is also ugly, I agree.

Since SRE is available inside the mj-single module, how about adding another export mj-single that simply makes the are routine available rather than passing it in the result data object?  You could add something like

```
module.exports.pprintXML = function (xml) {speech.pprintXML(xml)};
```

or whatever the right syntax is for that.  Then the `tex2stree` function can call `mjAPI.pprintXML()` when `--min` is specified.

Also, shouldn't there also be `asciimath2stree` and `mml2stree` as well?

bin/tex2stree
```diff
@@ -45,12 +49,14 @@ mjAPI.start();
...
   math: argv._[0],
   format: "TeX",
-  semantic: true
+  semantic: true,
+  minSTree: argv.min
 }, function (data) {
   if (!data.errors) {
     if (argv.xml) {
-      console.log(data.streeXml);}
+      console.log(argv.min ? data.streeXml : (data.streeXml));}
```
dpvc: shouldn't this just be the original line again, since you are returning `data.streeXML` in both cases?

dpvc:
I guess an alternative for v3.0 would be to make the returned object be an actual instance of a class that has a well-deinfed set of methods (like one to pretty print the result).  It could even be that ALL the conversion os done by methods.  E.g., you could use something like `var math = MathJax.TeX("\sqrt{x+1}")` to get an object from TeX input, and then `math.mml()` to get the MathML version, `math.svg()` to get the svg, `math.speech()` for the speech, etc.  If you wanted to use setters and getters, then you could even just do `math.mml` rather than `math.mml()`.  Something to think about for the future.

bin/tex2stree
zorkow: Absolutely! The buffer is still in my emacs and somehow I must have done some undo move before committing. Thanks for spotting that.

zorkow:
> Also, shouldn't there also be asciimath2stree and mml2stree as well?

Indeed, I wanted to add the former when this PR is done. 
For the latter one could use SRE directly. However, for completeness it should probably be here as well.

pkra:
> Also, shouldn't there also be asciimath2stree and mml2stree as well?

FWIW, I'm not a huge fan of the increasing number of examples in `bin`. They are slowly becoming a burden, I think.

dpvc:
> I'm not a huge fan of the increasing number of examples in `bin`

While the `bin` directory was originally intended as a set of examples, somewhere along the line they became real tools that people use.  So there is a conflict between keeping the number of examples small and providing the tools that people need.  As a set of examples, you are right, we don't need the other two commands, but as a set of tools, it makes sense to include them.  Perhaps the two functions need to be separated?  Or perhaps the tools need to be combined so that only one tool produces all the output forms, with a command line option that tells what to produce?  Or even one command for all input forms and output forms, with command line options that tell which? Or perhaps it could be based on the name of the command, so that the same code is used for all of them and what you name it determines the defaults?

pkra:
> So there is a conflict between keeping the number of examples small and providing the tools that people need. 

Yes. I worry we don't have the resources to maintain all the tools. The natural place for these would be separate modules, published to npm independently and maintained separately -- perhaps even community maintained. I admit I'm biased because I don't use them (except to test them and that seems a waste of energy). Do you use these in larger projects?

zorkow:
Alternatively, we could combine them all into a single tool, with many options. Much like pandoc for example.

dpvc:
> Do you use these in larger projects?

I don't have time for larger products; MathJax takes pretty much all my free time.  :smile:

dpvc:
> we could combine them all into a single tool, with many options

That was one of my suggestions as well.  Probably the one that requires the least code.

dpvc:
> The natural place for these would be separate modules, published to npm independently and maintained separately -- perhaps even community maintained.

Do you mean a separate module for each command?  Seems a bit excessive.  I could see the whole collection in one package, however.

zorkow:
Can this be merged now?
I'd like to get back to the MathJS guys soon.

zorkow:
Great! Thanks.

PR #194, opened 2016-03-12
Complete sre scripts

zorkow:
Adds the scripts to translate MathML and AsciiMath to semantic tree.

Further to our previous conversation: I really believe we should have a single binary with tons of options here, because I feel this is what people will (want to) use, instead of writing their own little wrappers.

### DIFF
--- a/bin/am2stree
+++ b/bin/am2stree
@@ -1,0 +1,62 @@
+#! /usr/bin/env node
+
+/*************************************************************************
+ *
+ *  am2stree
+ *
+ *  Uses MathJax to convert a AsciiMath string to SRE's semantic tree.
+ *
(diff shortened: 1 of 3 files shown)


dpvc:
> I really believe we should have a single binary with tons of options 

I'm fine with that, if we are providing actual tools for use.  As Peter mentioned, these were originally just examples of using the interface, and we expected that people would incorporate mathjax-node into larger projects via calls to the API (e.g., as Mathoid does), not via command-line tools.  As we move toward v3.0, I agree, the command-line tools should be consolidated.  I'm not sure it's worth the effort now, but it might not be all that hard, so maybe it is.

In any case, I'll merge.

PR #203, opened 2016-03-29
Fix exit codes

jcalvo-ixxus:
Sorry, I really wanted to merge it in develop branch... I just added exit codes to the commands used, if there is any error, so it doesn't return exitcode 0. There are some scripts which I didn't modify due to the complexity of the logic after MathJax execution.


pkra:
PRs need to go into develop (and this would also merge develop into master which is not in line with our branching model).

But since you write that this is incomplete, please file a bug report first.

PR #210, opened 2016-04-11
SVG+SRE: change role from math to img

### LINKED ISSUE #209: SVG+SRE sets role=math
This happens at https://github.com/mathjax/MathJax-node/blob/master/lib/mj-single.js#L577 .

We had changed this upstream to `role="img"` so I think we should match (cf. also https://github.com/mathjax/MathJax/issues/1437 about avoiding `role="math"`).

pkra:
Fixes #209

### DIFF
--- a/lib/mj-page.js
+++ b/lib/mj-page.js
@@ -746,7 +746,7 @@ function GetSpeech() {
       return queue.Push(err.restart,window.Array(SPEAK,svg));
     }
     jax.speech = speech.toSpeech(mml);
-    svg.setAttribute("role","math");
+    svg.setAttribute("role","img");
     svg.setAttribute("aria-labelledby",id+"-Title "+id+"-Desc");
     for (var i = 0, m = svg.childNodes.length; i < m; i++)
       svg.childNodes[i].setAttribute("aria-hidden",true);
--- a/lib/mj-page.js
+++ b/lib/mj-page.js
@@ -275,6 +275,7 @@ function ConfigureMathJax() {
                 {prev.parentNode.removeChild(prev)}
               //
               //  Add the span, and a div if in display mode,
+              //  then mark it as being processed
               //
               jax = script.MathJax.elementJax; if (!jax) continue;
               jax.SVG = {display: (jax.root.Get("display") === "block")}
@@ -398,9 +399,7 @@ function ConfigureMathJax() {
                 var NODE = CHTML.Element("mjx-chtml",{className:"MJXc-display",isMathJax:false});
                 NODE.appendChild(node); node = NODE;
               }
-              //
-              //  Mark math for screen readers
-              //
+
               node.className += " MJXc-processing";
               script.parentNode.insertBefore(node,script);
               //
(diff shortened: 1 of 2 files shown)


pkra:
There are other comments on setting the role to math. I'm not sure in how far those comments reflect the behavior right now.

dpvc:
The `img` role is now being added by MathJax already, so these lines can be removed.  Also, the comments about roles.

pkra:
Ah! I'll just drop them then.

pkra:
Updated.

pkra:
This way, this also doesn't need a test, I think (since it's now relying on upstream testing).

dpvc:
Can you also remove the "set role" part of the comments just above those lines?

pkra:
> Can you also remove the "set role" part of the comments just above those lines?

I'm not sure what you are referring to. Could you point me to a line in https://github.com/mathjax/MathJax-node/blob/issue209/lib/mj-page.js  or https://github.com/mathjax/MathJax-node/blob/issue209/lib/mj-single.js?

dpvc:
See [line 278](https://github.com/mathjax/MathJax-node/blob/issue209/lib/mj-page.js#L278) and [line 404](https://github.com/mathjax/MathJax-node/blob/issue209/lib/mj-page.js#L404) in mj-page and [line 277](https://github.com/mathjax/MathJax-node/blob/issue209/lib/mj-single.js#L277) and [line 403](https://github.com/mathjax/MathJax-node/blob/issue209/lib/mj-single.js#L403) in mj-single.

pkra:
Thanks!

lib/mj-page.js
```diff
@@ -275,7 +275,6 @@ function ConfigureMathJax() {
                 {prev.parentNode.removeChild(prev)}
               //
               //  Add the span, and a div if in display mode,
-              //  then set the role and mark it as being processed
```
dpvc: You took a bit too much.  The "and mark it as being processed" should have been kept.  I suppose it is not critical, but then the comma at the end of the precious line should be removed.

lib/mj-page.js
```diff
@@ -401,7 +400,6 @@ function ConfigureMathJax() {
               }
               //
               //  Mark math for screen readers
-              //    (screen readers don't know about role="math" yet, so use "textbox" instead)
```
dpvc: The "Mark math for screen readers" should be removed as well, since that is no longer being done.  So this whole comment can go.

lib/mj-single.js
```diff
@@ -274,7 +274,6 @@ function ConfigureMathJax() {
                 {prev.parentNode.removeChild(prev)}
               //
               //  Add the span, and a div if in display mode,
-              //  then set the role and mark it as being processed
               //
```
dpvc: Same as above.

lib/mj-single.js
```diff
@@ -400,7 +399,6 @@ function ConfigureMathJax() {
               }
               //
               //  Mark math for screen readers
-              //    (screen readers don't know about role="math" yet, so use "textbox" instead)
```
dpvc: Same as above.

pkra:
Thanks for the comments -- sorry for messing that up a bit. I hope it's better now.

pkra:
Oh, I had meant to try the new squashing commits on this one. Oh well, I'm sure I'll screw another one up so I can try that :wink:

dpvc:
Sorry!  I was trying to be responsive and actually merge some pull requests!

PR #211, opened 2016-04-12
[mj-single] add speech to CommonHTML output

### LINKED ISSUE #189: Add speech output to CommonHTML output
We should add the speech output to the HTML output if requested.

pkra:
Fixes #189.


pkra:
I'm not sure about trying to fix mj-page. Any thoughts?

dpvc:
> I'm not sure about trying to fix mj-page. Any thoughts?

If you are going to decommission mj-page, then I guess it is not worth it.  It does look like it would take some refactoring.

Issue #206, opened 2016-04-11
move mj-page out of mathjax-node


pkra:
As per F2F (cf #191), we will move `mj-page` out of mathjax-node; there will be alternative modules to use.

pkra:
Here's the first draft for a separate module based on svg2png

```javascript
var mjAPI = require("mathjax-node/lib/mj-single.js");

exports.math2png = function(mjoptions, pngoptions, callback){
  mjAPI.start();
  var svg2png = function(result, options, callback){
    if (result.errors) return result.errors;
    var svgpng = require('svg2png');
    var sourceBuffer = new Buffer(result.svg, "utf-8");
    var returnBuffer = svgpng.sync(sourceBuffer, {
      width: result.svg.width,
      height: result.svg.height
    });
    result.png = {};
    result.png.data = "data:image/png;base64," + returnBuffer.toString('base64');
    // maybe have something for scaling / dpi?
    return callback(result);
  };
  // make sure SVG output will be generated
  mjoptions.svg = true;
  return mjAPI.typeset(mjoptions, function(result){
    console.log();
    svg2png(result, pngoptions, callback);
  });
}
```

dpvc:
A couple of comments:

1.  This doesn't use `mjAPI.config()`, and that means you can't control the font, for example.  You don't necessarily have to allow full access to `mjAPI.config()`, but allowing configuration of fonts is probably a good idea.  I suppose the application could itself require mj-single and call `mjAPI.config()` itself, but perhaps that needs to be encompassed in another (configuration) module?  I haven't really thought much about how these modules should be configured.

2.  That brings up the question of the `mjoptions` and `pngoptions` arguments.  As it stands the `mjoptions` could specify creation of HTML or MathML output, and those formats would be generated and discarded (inefficient).  Perhaps you want to make your own option object for the `typeset()` call and copy the options that make sense from the `mjoptions` object rather than passing them all in unfiltered.

    Also, the `pngoptions` are not used anywhere (I assume these are for future use).  I'm wondering if there need to be two sets of options.  If you create your own options for mathjax from the ones passed in, you could dal so create your own options for the png conversion from one option object that the user passes to `math2png`.

[...]

PR #213, opened 2016-04-15
Remove support for PNG generation

### LINKED ISSUE #205: move PNG generation out of mathjax-node
As per F2F, the PNG generation will be moved out of mathjax-node; there will be alternative modules to use.

Cf. the discussion in #174; see also #191.

pkra:
Fixes #205.


pkra:
Should we make another release before v1.0 and other breaking changes (like #206)?

pkra:
This is marked WIP because of the open questions in #205.

pkra:
The only remaining question I had was whether https://github.com/mathjax/MathJax-node/blob/master/lib/mj-single.js#L587-L593 could/should be dropped or not. 

But that's a minor item I think (and could be done when #219 is decided) so I think this is ready to go for now.

pkra:
Ahem, this is not quite ready. I need to remove all other PNG-related things (readme, examples in bin etc)

pkra:
Ok, that's better. There are still some references in mj-page but that's for #206 to fix.

README.md
```diff
@@ -18,10 +18,7 @@ Use
 
 to install mathjax-node and its dependencies.
 
-These API's can produce PNG images, but that requires the
-[Batik](http://xmlgraphics.apache.org/batik/download.html) library.  It 
-should be installed in the `batik` directory.  See the README file in that 
-directory for more details.
+These API's no longer produce PNG images but can easily be integrated into workflows using svg-to-png converters (e.g., png2svg, librsvg, batik).
```
dpvc: Do you really mean `png2svg` here?  That sounds like the wrong direction.

dpvc:
Just to be clear, this also removes not only the PNG generation, but the ability to produce image tags that use remote SVG files as well.  I assume you want to move that to a post-processing phase, but wanted to make sure that it was part of the plan, since it has not been discussed anywhere else and isn't in the PR title.

README.md
pkra: Thanks for catching that.

pkra:
> Just to be clear, this also removes not only the PNG generation, but the ability to produce image tags that use remote SVG files as well. I assume you want to move that to a post-processing phase, but wanted to make sure that it was part of the plan, since it has not been discussed anywhere else and isn't in the PR title.

Right. I actually remember not being able to figure out if this code was still relevant but then deciding that it also fits more naturally in post-processing. I can pull out the commit if you prefer (I have to rebase anyway).

dpvc:
>  I can pull out the commit if you prefer

No need.  Leave it in, I just wanted it to be in the record.

pkra:
@dpvc I rebased (and addressed your comment). I'm a little insecure as to the rebase (lots of changes within mj-single) but I went through it twice and I didn't spot any bad changes from the resolution.

dpvc:
There rebase looks good.  I'll merge.

PR #216, opened 2016-04-19
[mj-single] HTML output: remove MathJax ids and element frame

### LINKED ISSUE #215: [mj-single] The HTML output should not include id attributes
The HTML output should probably have its id's removed (those that were added by the CommonHTML output).  Otherwise, we will end up with duplicate id's if the output from two conversions are included in the same page.

Also, the CHTML id's may want to be reset every typeset so that (if used as the backend of some other service) the id's don't grow unbounded.

Also, the `-Frame` element could be removed, as it is for the other output formats.

pkra:
Fixes #215

### DIFF
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -570,7 +570,22 @@ function GetHTML(result) {
     for (var i=0, m=html.childNodes.length; i < m; i++)
       html.childNodes[i].setAttribute("aria-hidden",true);
   }
-
+  // remove automatically generated IDs
+  var ids = html.querySelectorAll('[id^="MJXc-Node-"]');
+  for (var i = 0; i < ids.length; i++){
+    ids[i].removeAttribute("id");
+  }
+  // remove extreneous frame element
+  var frame = html.querySelector('[id^="MathJax-Element-"]');
+  if (frame){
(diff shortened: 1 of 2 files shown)


lib/mj-single.js
```diff
@@ -570,7 +570,22 @@ function GetHTML(result) {
...
+  // remove extreneous frame element
+  var frame = html.querySelector('[id^="MathJax-Element-"]');
+  if (frame){
+    // in display-mode, the frame is inside the display-style wrapper
+    html.insertBefore(frame.firstChild, frame);
+    html.removeChild(frame);
+  }
+  else{
+    // otherwise (inline-mode) the frame is the root element
+    html.removeAttribute("id");
```
pkra: I wasn't sure if the `class=mjx-chtml` on the wrapper and -frame could just be moved to the element with `class=mjx-math (similarly for the `MJXc-display` in display-mode).

dpvc: They probably can, but I haven't done any testing to check for sure.  Personally, I'd leave the outer display element rather than move the class to the inner one, but it would probably still work if you did.

pkra: Copy that. I'm leaving it as it is.

pkra:
I think this is ready.

PR #227, opened 2016-05-02
Use <script> tags rather than delimited math in mj-single.

### LINKED ISSUE #226: extraneous delimiters cause unexpected results
While responding to [@fptoth's comment](https://github.com/mathjax/MathJax-node/issues/206#issuecomment-216064704), I noticed that 

```latex 
$$  VBeBorn=reflexiveVP\ (P.dirV2(v\_besch106{\hbox{}}na\c{s}te{\hbox{"}})).$$
```

passed to `mj-single` would yield an empty result (e.g., `<math xmlns="http://www.w3.org/1998/Math/MathML" display="block" />` etc.)

I'm not sure why the incorrect delimiters would throw mathjax-node off.

dpvc:
This resolves the problem of having delimiters as part of the math passed to mj-single, and also prevents the need for the preprocessors.

[Fixes #226]

### DIFF
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -106,7 +106,7 @@ var ID = 0;           // id for this SVG element
 //  The delimiters used for each of the input formats
 //
 var TYPES = {
-  TeX: "tex;display=block",
+  TeX: "tex; mode=display",
   "inline-TeX": "tex",
   AsciiMath: "asciimath",
   MathML: "mml"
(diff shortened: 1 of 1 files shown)


dpvc:
OK, the changes fail when the math is empty.  I'll have to fix that...

dpvc:
OK, the problem was a `<` where there should have been a `>` to end a tag.  OOPS!

Also, I had the wrong display-math mercer for the TeX script tag.  The tests should pass now.

PR #229, opened 2016-05-03
Define outerEm for CHTML.

### LINKED ISSUE #223: [CommonHTML] displayIndent not working
I couldn't get displayIndent to work.  Hopefully with a correct configuration this time:

```javascript 
var tape = require('tape');
var mjAPI = require("./lib/mj-single.js");
var jsdom = require('jsdom').jsdom;

tape('displayAlign:left in HTML output', function(t) {
  t.plan(1);
  mjAPI.config({MathJax: {displayIndent: '10em'}});
  mjAPI.start();
  var tex = 'x';
  expected = -1;

  mjAPI.typeset({
    math: tex,
    format: "TeX",
    html: true
  }, function(data) {
    var document = jsdom(data.html);
    var window = document.defaultView;
    var element = window.document.getElementsByClassName('MJXc-display')[0];
    var result = element.getAttribute('style').indexOf('margin-left');
console.log(result);
    t.ok(result > expected, 'style includes a margin');
  });
});
```
Perhaps another jsdom bug (like #220)?

pkra:
Fixes #223.


pkra:
@dpvc what would a good test look like? (The test I had included in the op of #223 passes on this branch now but you wrote on the issue that it's testing the wrong thing.)

dpvc:
> what would a good test look like? (The test I had included in the op of #223 passes on this branch now but you wrote on the issue that it's testing the wrong thing.)

The test is fine for the equation used.  I was thinking of the case where there is an equation label (tagged equation), in which case the `left` is on a different element.  I don't know if you want a test for that case as well.

pkra:
Ok, I'll add the test and another one for labeled equations.

pkra:
I've added two tests (labeled and unlabeled cases).

I think this is ready.

test/issue223-1.js
```diff
@@ -0,0 +1,26 @@
...
+    math: tex,
+    format: "TeX",
+    html: true
+  }, function(data) {
+    console.log(data.html);
+    var document = jsdom(data.html);
+    var window = document.defaultView;
+    var element = window.document.getElementsByClassName('MJXc-display')[0];
+    var result = element.getAttribute('style').indexOf('margin-left');
+    console.log(result);
```
dpvc: Do you really want the `console.log()` calls here and at line 18?

test/issue223-1.js
```diff
@@ -0,0 +1,26 @@
...
+  mjAPI.typeset({
+    math: tex,
+    format: "TeX",
+    html: true
+  }, function(data) {
+    console.log(data.html);
+    var document = jsdom(data.html);
+    var window = document.defaultView;
+    var element = window.document.getElementsByClassName('MJXc-display')[0];
+    var result = element.getAttribute('style').indexOf('margin-left');
```
dpvc: This test seems a bit odd.  Why not 

```
  var result = element.style.marginLeft;
```

or 

```
  var result = (element.style.marginLeft === "");
```

rather than doing a string search on the styles attribute?

test/issue223-1.js
```diff
@@ -0,0 +1,26 @@
+
+var tape = require('tape');
+var mjAPI = require("../lib/mj-single.js");
+var jsdom = require('jsdom').jsdom;
+
+tape('displayAlign:left in HTML output', function(t) {
```
dpvc: Your test says it is about `displayAlign`, but seems to be about `displayIndent` according to the `config()` call below.

test/issue223-2.js
```diff
@@ -0,0 +1,27 @@
...
+var tape = require('tape');
+var mjAPI = require("../lib/mj-single.js");
+var jsdom = require('jsdom').jsdom;
+
+tape('displayAlign:left in HTML output with mlabeledtr', function(t) {
+  t.plan(1);
+  mjAPI.config({MathJax: {displayIndent: '10em'}
+});
+  mjAPI.start();
+  var tex = '\\begin{equation} x \\tag{1} \\end{equation}';
```
dpvc: There is no need for the `\\begin{equation}` and `\\end{equation}` here; `x \\tag{1}` should be sufficient.

test/issue223-2.js
```diff
@@ -0,0 +1,27 @@
...
+    math: tex,
+    format: "TeX",
+    html: true
+  }, function(data) {
+    console.log(data.html);
+    var document = jsdom(data.html);
+    var window = document.defaultView;
+    var element = window.document.getElementsByClassName('mjx-table')[0];
+    var result = element.getAttribute('style').indexOf('margin-left');
+    console.log(result);
```
dpvc: Same questions here as above.

pkra:
Thanks, @dpvc. I've fixed (and merged) the tests. Hope this is ready now.

PR #232, opened 2016-05-03
[mj-single] [CommonHTML] reset internal IDs after processing

### LINKED ISSUE #231: [mj-single] [CHTML] reset internal ids
Split out of https://github.com/mathjax/MathJax-node/issues/215#issue-148969074:

> CHTML id's may want to be reset every typeset so that (if used as the backend of some other service) the id's don't grow unbounded.
> ...
> So something like `MathJax.OutputJax.CommonHTML.ID = 0;` should do (untested).

pkra:
Fixes #231

### DIFF
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -760,6 +760,7 @@ function GetState(state) {
     }
     if (TEX.resetEquationNumbers) TEX.resetEquationNumbers();
     MML.SUPER.ID = ID = 0;
+    MathJax.OutputJax.CommonHTML.ID = 0;
   }
 }
 


pkra:
FWIW something like the test below tests this PR -- but only until #215 is resolved (cf. my comment on #231)

```javscript
var tape = require('tape');
var mjAPI = require("..//lib/mj-single.js");
var jsdom = require('jsdom').jsdom;

tape('Reset automatically generated IDs for HTML output', function(t) {
  t.plan(1);

  mjAPI.start();
  var expected = 'MathJax-Element-1-Frame';

  mjAPI.typeset({
    math: "x",
    format: "TeX",
    html: true
  }, function(data) {
    console.log(data.html);
  });
  mjAPI.typeset({
    math: "y",
    format: "TeX",
    html: true
  }, function(data) {
    console.log(data.html);
    var document = jsdom(data.html);
    var window = document.defaultView;
    var id = window.document.querySelector('[id^="MathJax-Element-"]').getAttribute('id');
    t.equal(id, expected, 'automatic ids successfully reset');
  });
});
```

dpvc:
Once IDs are removed, I don't see any way to test this.  It would require access to the MathJax running in the jsdom window, but the module structure doesn't allow for that.  It would be nice to be able to use the MathJax `AuthorInit` function, but `mj-single` already uses that, so setting it would disable mathjax-node.  I suppose we could modify `mj-single` to handle `AuthorInit` from the user configuration so that we could insert the code to test the IDs.

pkra:
> Once IDs are removed, I don't see any way to test this. 

Copy that.

>  I suppose we could modify mj-single to handle AuthorInit from the user configuration so that we could insert the code to test the IDs.

Probably not worth it before 3.0 comes around. Since it's mostly a convenience improvement, I'm ok with not having a test.

Issue #233, opened 2016-05-17
[SVG output] margin-bottom lost


pkra:
Via https://phabricator.wikimedia.org/T135423; cc @physikerwelt.

It seems like margin-bottom is missing when generating output with mathjax-node.

E.g., `$\underline{\mathbb{Z}}$` will get `style="vertical-align: -0.52ex;"` from mathjax-node but in a page would get `vertical-align: -0.52ex; margin-bottom: -0.734ex;`.

dpvc:
It looks like the CSSStyleDeclaration (used by jsdom) handling of `ex` units isn't right, so the `margin-bottom` isn't being set.  I'll have to look into it.  It is part of the code I had supplied to them, but it used their routines for determining when a measurement is valid.

dpvc:
Looks like their unit parser doesn't include `ex`.  Argh!  I'll have to see about patching that.

pkra:
Thanks for looking into it, @dpvc.

dpvc:
OK, I've included a patch in the `issue233` branch.  I still have to make a PR to the CSSStyleDeclaration site to fix the issue, but the patch should take care of it from our end for now.

dpvc:
Note, the misalignment in the original complaint is due to the fact that method adjusts the placement (I haven't looked into how it does this), which may be necessary because it is a background image rather than a direct image element.  It may be that the missing `margin-bottom` was causing their computations to go awry.

physikerwelt:
@dpvc Thank you very much. The patch seems to work great. However, I have to admit that I do not understand how it works. Since quite a few lines were changed, I'm a afraid that it might break other things.

physikerwelt:
OK. Ater having understood that https://github.com/mathjax/MathJax-node/commit/bd1d6c430dbd45d5e6db8e496bfc03619e411169#diff-c4f7e206b09703de4edb0800d5a5af15R386 really is a non trivial setter and the line below is a getter I can finally get an idea of the change.

PR #234, opened 2016-05-17
Patch CSSStyleDeclaration to handle units of 'ex' (and others).  Resolves issue #233

dpvc:
This patches the copy of CSSStyleDeclaration to handle units of `ex` and a number of other missing units.  The change really should only be in one line (the `lengthRegEx`), but since that is down inside a closure, I can't get my hands on it, so I have to replace the entire `valueType()` routine with a duplicate that uses my own copy of `lengthRegEx`.  Resolves issue #233.

### DIFF
--- a/lib/patch/jsdom.js
+++ b/lib/patch/jsdom.js
@@ -171,6 +171,113 @@ var WIDTH = (function () {
 
 
 //
+//  Patch for CSSStyleDeclaration lengthRegEx so that it includes ex units
+//  (plus a number of other units that are left out)
+//  
+var FixValueType = function () {
+  var parsers = require(PARSERS);
+  
+  var integerRegEx = /^[\-+]?[0-9]+$/;
+  var numberRegEx = /^[\-+]?[0-9]*\.[0-9]+$/;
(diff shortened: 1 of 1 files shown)


pkra:
Travis fails one test -- but the basic SRE one. @zorkow could you check that out?

@dpvc could this use a test based on the underline example?

dpvc:
> could this use a test based on the underline example?

Since it is a jsdom (or rather CSSStyleDeclaration) problem, not a mathjax-node problem, I'm not sure it needs a test here.  If you want some tests for the various jsdom issues that we have patched, then I suppose that could be done, but I think it is not necessary.

dpvc:
PS, the tests passed when I ran them by hand, but I had an older copy of SRE.  When I update that, the tests fail.  :-(

pkra:
> Since it is a jsdom (or rather CSSStyleDeclaration) problem, not a mathjax-node problem, I'm not sure it needs a test here.

Sounds good to me.

pkra:
The tests now pass thanks to @zorkow's fixes to SRE. Yay to travis for catching these.

Issue #207, opened 2016-04-11
move SRE integration out of mj-node


pkra:
As per F2F (cf. #191), the SRE integration will be moved out of mathjax-node; there will be alternative modules.

zorkow:
Just started looking into this. I assume it's best tackled after PR #247 is merged.

pkra:
I agree -- best to wait until #247.

pkra:
From F2F, we should preserve the code in mj-single to add dummy speech text (e.g., TeX/asciimath source); this will be helpful for post-processing, in particular for the SVG infrastructure of labeledby/title (running IDs etc).

pkra:
Merged into `develop`.

PR #237, opened 2016-05-25
fix speakText + original alttext leading to duplication in output

### LINKED ISSUE #236: [speech] alttext on MathML input leads to duplication in output
[Splitting this from https://github.com/mathjax/MathJax-node/issues/121#issuecomment-190666132]

If MathML input has `alttext` and mathjax-node is set to generate output with speechtext, then we get a duplication.

For MathML, the source alttext is changed to the speechtext but the attribute is duplicated in the output. This is due to a check missing at https://github.com/mathjax/MathJax-node/blob/master/lib/mj-single.js#L544

For HTML, the speechtext is added as an aria-label on the outer span; the source alttext is added to its first child (but that's in an aria-hidden element). This is due to https://github.com/mathjax/MathJax-node/blob/develop/lib/mj-single.js#L570-L574

For SVG output, the source `alttext` ends up on an aria-label (because of https://github.com/mathjax/MathJax/blob/develop/unpacked/jax/output/SVG/jax.js#L2111) and the speechtext ends up in a `title` tag pointed to by an `aria-labeledby` (due to https://github.com/mathjax/MathJax-node/blob/master/lib/mj-single.js#L575-L585).

Examples;

### MML

```bash 
$ node bin/mml2mml --speech '<math alttext="hello world"> <msup> <mi>&#x03B1;<!-- α --></mi> 
```
produces 

```html
<mn>2</mn> </msup> </math>'
<math xmlns="http://www.w3.org/1998/Math/MathML" alttext="alpha squared" alttext="alpha squared">
  <msup>
    <mi>&#x3B1;</mi>
    <mn>2</mn>
  </msup>
</math>
```

### HTML

```bash
$ node bin/mml2html --speech '<math alttext="hello world"> <msup> <mi>&#x03B1;<!-- α --></mi> <mn>2</mn> </msup> </math>'
```

produces 

```html
<span class="mjx-chtml" aria-label="alpha squared"><span class="mjx-math" aria-label="hello world" role="math" aria-hidden="true"><span class="mjx-mrow"><span class="mjx-msup"><span class="mjx-base"><span class="mjx-mi"><span class="mjx-char MJXc-TeX-math-I" style="padding-top: 0.225em; padding-bottom: 0.298em;">α</span></span></span><span class="mjx-sup" style="font-size: 70.7%; vertical-align: 0.513em; padding-left: 0px; padding-right: 0.071em;"><span class="mjx-mn" style=""><span class="mjx-char MJXc-TeX-main-R" style="padding-top: 0.372em; padding-bottom: 0.372em;">2</span></span></span></span></span></span></span>
```

### SVG

```bash
$ node bin/mml2svg --speech '<math alttext="hello world"> <msup> <mi>&#x03B1;<!-- α --></mi><mn>2</mn> </msup> </math>'
```

will produce

```html
<svg xmlns:xlink="http://www.w3.org/1999/xlink" width="2.542ex" height="2.509ex" style="vertical-align: -0.338ex;" viewBox="0 -934.9 1094.4 1080.4" aria-label="hello world" role="math" focusable="false" xmlns="http://www.w3.org/2000/svg" aria-labelledby="MathJax-SVG-1-Title MathJax-SVG-1-Desc">
<title id="MathJax-SVG-1-Title">Equation</title>
<desc id="MathJax-SVG-1-Desc">alpha squared</desc>
<defs aria-hidden="true">
<path stroke-width="1" id="E1-MJMATHI-3B1" d="M34 156Q34 270 120 356T309 442Q379 442 421 402T478 304Q484 275 485 237V208Q534 282 560 374Q564 388 566 390T582 393Q603 393 603 385Q603 376 594 346T558 261T497 161L486 147L487 123Q489 67 495 47T514 26Q528 28 540 37T557 60Q559 67 562 68T577 70Q597 70 597 62Q597 56 591 43Q579 19 556 5T512 -10H505Q438 -10 414 62L411 69L400 61Q390 53 370 41T325 18T267 -2T203 -11Q124 -11 79 39T34 156ZM208 26Q257 26 306 47T379 90L403 112Q401 255 396 290Q382 405 304 405Q235 405 183 332Q156 292 139 224T121 120Q121 71 146 49T208 26Z"></path>
<path stroke-width="1" id="E1-MJMAIN-32" d="M109 429Q82 429 66 447T50 491Q50 562 103 614T235 666Q326 666 387 610T449 465Q449 422 429 383T381 315T301 241Q265 210 201 149L142 93L218 92Q375 92 385 97Q392 99 409 186V189H449V186Q448 183 436 95T421 3V0H50V19V31Q50 38 56 46T86 81Q115 113 136 137Q145 147 170 174T204 211T233 244T261 278T284 308T305 340T320 369T333 401T340 431T343 464Q343 527 309 573T212 619Q179 619 154 602T119 569T109 550Q109 549 114 549Q132 549 151 535T170 489Q170 464 154 447T109 429Z"></path>
</defs>
<g stroke="currentColor" fill="currentColor" stroke-width="0" transform="matrix(1 0 0 -1 0 0)" aria-hidden="true">
 <use xlink:href="#E1-MJMATHI-3B1" x="0" y="0"></use>
 <use transform="scale(0.707)" xlink:href="#E1-MJMAIN-32" x="905" y="513"></use>
</g>
</svg>
```

pkra:
Fixes #236

### DIFF
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -572,7 +572,8 @@ function GetHTML(result) {
 
   if (data.speakText) {
     var label = html.querySelector("[aria-label]");
-    if (label) label.removeAttribute("aria-label");    html.setAttribute("aria-label",result.speakText);
+    if (label) label.removeAttribute("aria-label");
+    html.setAttribute("aria-label",result.speakText);
     for (var i=0, m=html.childNodes.length; i < m; i++)
       html.childNodes[i].setAttribute("aria-hidden",true);
   }
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -571,6 +571,7 @@ function GetHTML(result) {
   var script = jax.SourceElement(), html = script.previousSibling;
 
   if (data.speakText) {
+    if(html.querySelector("[aria-label]")){html.querySelector("[aria-label]").removeAttribute("aria-label");}
     html.setAttribute("aria-label",result.speakText);
     for (var i=0, m=html.childNodes.length; i < m; i++)
       html.childNodes[i].setAttribute("aria-hidden",true);
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -543,7 +543,10 @@ function GetSpeech(result) {
   if (!data.mml) return;
   var jax = MathJax.Hub.getAllJax()[0];
   jax.root.alttext = result.speakText;
-  if (jax.root.attrNames) {jax.root.attrNames.push("alttext")}
+  var attrNames = jax.root.attrNames;
+  if (attrNames & (attrNames.indexOf("alttext") == -1)) {
+    attrNames.push("alttext");
+  }
   result.mml = jax.root.toMathML('',jax);
 }
 
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -571,8 +571,8 @@ function GetHTML(result) {
   var script = jax.SourceElement(), html = script.previousSibling;
 
   if (data.speakText) {
-    if (html.querySelector("[aria-label]")){ html.querySelector("[aria-label]").removeAttribute("aria-label");}
-    html.setAttribute("aria-label",result.speakText);
+    var label = html.querySelector("[aria-label]");
+    if (label) label.removeAttribute("aria-label");    html.setAttribute("aria-label",result.speakText);
     for (var i=0, m=html.childNodes.length; i < m; i++)
       html.childNodes[i].setAttribute("aria-hidden",true);
   }
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -610,6 +610,7 @@ function GetSVG(result) {
   //
   if (data.speakText) {
     ID++; var id = "MathJax-SVG-"+ID;
+    svg.removeAttribute("aria-label");
     svg.setAttribute("aria-labelledby",id+"-Title "+id+"-Desc");
     for (var i=0, m=svg.childNodes.length; i < m; i++)
       svg.childNodes[i].setAttribute("aria-hidden",true);
--- a/lib/mj-single.js
+++ b/lib/mj-single.js
@@ -544,7 +544,7 @@ function GetSpeech(result) {
   var jax = MathJax.Hub.getAllJax()[0];
   jax.root.alttext = result.speakText;
   var attrNames = jax.root.attrNames;
-  if (attrNames & (attrNames.indexOf("alttext") == -1)) {
+  if (attrNames && attrNames.indexOf("alttext") === -1) {
     attrNames.push("alttext");
   }
   result.mml = jax.root.toMathML('',jax);
@@ -571,7 +571,7 @@ function GetHTML(result) {
   var script = jax.SourceElement(), html = script.previousSibling;
 
   if (data.speakText) {
-    if(html.querySelector("[aria-label]")){html.querySelector("[aria-label]").removeAttribute("aria-label");}
+    if (html.querySelector("[aria-label]")){ html.querySelector("[aria-label]").removeAttribute("aria-label");}
     html.setAttribute("aria-label",result.speakText);
     for (var i=0, m=html.childNodes.length; i < m; i++)
       html.childNodes[i].setAttribute("aria-hidden",true);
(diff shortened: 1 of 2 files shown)


pkra:
@zorkow travis fails because of the SRE test. Could you look into it?

lib/mj-single.js
```diff
@@ -543,7 +543,10 @@ function GetSpeech(result) {
   if (!data.mml) return;
   var jax = MathJax.Hub.getAllJax()[0];
   jax.root.alttext = result.speakText;
-  if (jax.root.attrNames) {jax.root.attrNames.push("alttext")}
+  var attrNames = jax.root.attrNames;
+  if (attrNames & (attrNames.indexOf("alttext") == -1)) {
```
dpvc: This should be `&&` not just `&` (which is a bitwise operation).  Then you will not need the extra parentheses around the equality check.  The equality should be `===` rather than `==`, since no type-coersion is needed.

lib/mj-single.js
```diff
@@ -568,6 +571,7 @@ function GetHTML(result) {
   var script = jax.SourceElement(), html = script.previousSibling;
 
   if (data.speakText) {
+    if(html.querySelector("[aria-label]")){html.querySelector("[aria-label]").removeAttribute("aria-label");}
```
dpvc: I'd like to see some spaces here (after `if`, before the `{`).

test/issue236.js
```diff
@@ -0,0 +1,44 @@
...
+    // test svg output
+    var document = jsdom(data.svg);
+    var window = document.defaultView;
+    var labeled = window.document.querySelector('[aria-label]');
+    t.notOk(labeled, 'svg output has no aria-label');
+
+    var labelledby = window.document.querySelector('[aria-labelledby]');
+    t.ok(labelledby, 'svg output has the correct aria-labelledby');
+
+    // text mml output
```
dpvc: I think this is "test" not "text"?

pkra:
Thanks, @dpvc. I've updated the PR.

lib/mj-single.js
```diff
@@ -568,6 +571,7 @@ function GetHTML(result) {
   var script = jax.SourceElement(), html = script.previousSibling;
 
   if (data.speakText) {
+    if (html.querySelector("[aria-label]")){ html.querySelector("[aria-label]").removeAttribute("aria-label");}
```
dpvc: I think it should be `("[aria-label]")) {html` not `("[aria-label]")){ html`, here.  Volker would probably like

```
if (html.querySelector("[aria-label]")) {
  html.querySelector("[aria-label]").removeAttribute("aria-label");
}
```
even better.  Or

```
var label = html.querySelector("[aria-label]");
if (label) label.removeAttribute("aria-label");
```

dpvc:
I'm not sure a decision was made about whether `alttext` should override speech text or *vice versa*.  I prefer keeping the `alttext` if there is one, but your comment about #207 makes sense, so I'm OK with this for now and will merge it once you have a chance to react to the comments I just left.

lib/mj-single.js
pkra: Thanks. I'll update the PR tomorrow.

pkra:
> I'm not sure a decision was made about whether alttext should override speech text or vice versa. I prefer keeping the alttext if there is one, but your comment about #207 makes sense, so I'm OK with this for now and will merge it once you have a chance to react to the comments I just left.

I thought I had brought it up at a F2F but I'm ok with switching. We can discuss it at the next F2F, too.

dpvc:
> I thought I had brought it up at a F2F but I'm ok with switching. We can discuss it at the next F2F, too.

I don't recall, but perhaps it is one that I missed.  We can discuss it tomorrow if you want.  As I said, it is not critical either way, as the speech generation is going be removed anyway.

pkra:
I've updated the PR.

lib/mj-single.js
```diff
@@ -571,8 +571,8 @@ function GetHTML(result) {
   var script = jax.SourceElement(), html = script.previousSibling;
 
   if (data.speakText) {
-    if (html.querySelector("[aria-label]")){ html.querySelector("[aria-label]").removeAttribute("aria-label");}
-    html.setAttribute("aria-label",result.speakText);
+    var label = html.querySelector("[aria-label]");
+    if (label) label.removeAttribute("aria-label");    html.setAttribute("aria-label",result.speakText);
```
dpvc: I think you left out a line break.  Sorry for the picky spacing issues!

pkra: Sigh.

pkra:
Hopefully without any more silly errors from my side. Sorry about that.

Issue #244, opened 2016-06-10
[CommonHTML] equation label not properly aligned in large arrays with many labels


pkra:
This is a continuation from https://github.com/mathjax/MathJax/issues/1500, I'm afraid.

The minimal example I had posted seems to be fixed in both MathJax and mathjax-node (where I had run into this). However, a more complex example (the original one where I encountered this) seems fixed in the browser, but still broken in mathjax-node.

Below is the (pretty messy) example.

```latex
\begin{eqnarray}
  && \mathfrak{d}\mathfrak{d}' =\delta _{\mathfrak{d}, \mathfrak{d}'} \mathfrak{d}. \tag{57}\cssId{i}{}{}\\
  && \mathfrak{d}' E_n \mathfrak{d}=0, \quad {\mathit{if}}\ {\mathrm{sgn}}(\mathfrak{d}') =\pm , \tag{58}\cssId{ii}{}{}\\
 && \mathfrak{d}' F_n \mathfrak{d}=0, \quad {\mathit{if}}\ {\mathrm{sgn}}(\mathfrak{d}) = \pm , \\
 && \mathfrak{d}E_h [(\mathfrak{d}- E^{\theta }_{h, h} + E^{\theta }_{h+1,h+1}, {\mathrm{sgn}}(\mathfrak{d}) ) ] = E_h [(\mathfrak{d}- E^{\theta }_{h, h} + E^{\theta }_{h+1,h+1},{\mathrm{sgn}}(\mathfrak{d}))], \\
 && \mathfrak{d}F_h [(\mathfrak{d}+ E^{\theta }_{h, h} - E^{\theta }_{h+1,h+1}, {\mathrm{sgn}}(\mathfrak{d}))]\\
&&\qquad \qquad \qquad = F_h [(\mathfrak{d}+ E^{\theta }_{h, h} - E^{\theta }_{h+1,h+1}, {\mathrm{sgn}}(\mathfrak{d}))],\quad {\mathit{if}} \ h\neq n, \\
  && F_nE_n \mathfrak{d}- \mathfrak{d}' F_n E_n \mathfrak{d}= [\lambda _n]_v \mathfrak{d}, \quad \quad \quad \hspace{58pt}\! {\mathit{if}} \ \lambda =\lambda ', \epsilon = - \epsilon ' \neq 0, \tag{59}\cssId{iii}{}{}\\
[...]

PR #245, opened 2016-06-10
Make sure padding and margin can accept 0 to mean 0px.  #244

dpvc:
Make sure padding and margin can accept 0 to mean 0px.  Resolves issue #244

### DIFF
--- a/lib/patch/jsdom.js
+++ b/lib/patch/jsdom.js
@@ -49,7 +49,7 @@ var PADDING = (function () {
 
   var isValid = function (v) {
     var type = parsers.valueType(v);
-    return type === TYPES.LENGTH || type === TYPES.PERCENT;
+    return type === TYPES.LENGTH || type === TYPES.PERCENT || type === TYPES.INTEGER;
   };
 
   var parser = function (v) {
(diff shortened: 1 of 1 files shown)


pkra:
I can't really comment on what it does but the it looks good to me.

pkra:
@dpvc I still see the issue (though it's not as bad as before). Example: http://codepen.io/pkra/pen/YWqRLB.

I'm using MathJax's develop branch and this branch.

dpvc:
> I still see the issue (though it's not as bad as before).

That turns out to be independent of this PR, so I'd recommend merging this for now.  The problem you are having is very dependent on the font size and zoom level (for me), and is probably due to rounding issues.  The labels are in a separate column that has spacing added to (theoretically) make it match the table row sizes, but the imprecise knowledge of the em and ex size affect this, as does the fact that the table is made of pieces with different heights than the labels, so the cumulative around-off error is different.  Also, the error is additive, so it gets more and more several as you go down the table.

I'm not sure how to fix it, but any solution will have to be in the CommonHTML code, not mathjax-node.  I'll see what I can come up with.

pkra:
> I'm not sure how to fix it, but any solution will have to be in the CommonHTML code, not mathjax-node. I'll see what I can come up with.

Will you open an issue upstream?

dpvc:
Done.  mathjax/MathJa#1532

PR #247, opened 2016-06-24
remove mj-page

### LINKED ISSUE #206: move mj-page out of mathjax-node

pkra:
Resolves #206


pkra:
@dpvc I started to write a simple [*2jax processor](https://gist.github.com/pkra/1ae75a4bb86fbc3364c8d85675030223) to add to `bin` as part of this PR.

I ran into two problems. First, the TeX input jax doesn't load (I get the 15s timeout and obviously this isn't terribly problematic). Second, when there is more than one input type on a page (i.e., TeX and ascii), then it hangs completely (or at least long enough for me to kill it).

If this approach make sense, do you have a suggestion for making it work?

test/base-mathjax.js
```diff
@@ -1,5 +1,5 @@
 var tape = require('tape');
-var mjAPI = require("..//lib/mj-single.js");
+var mjAPI = require("..//lib/main.js");
```
zorkow: Remove double slash.

test/base-speechruleengine.js
```diff
@@ -1,5 +1,5 @@
 var tape = require('tape');
-var mjAPI = require("..//lib/mj-single.js");
+var mjAPI = require("..//lib/main.js");
```
zorkow: ditto

test/issue104.js
```diff
@@ -1,5 +1,5 @@
 var tape = require('tape');
-var mjAPI = require("..//lib/mj-single.js");
+var mjAPI = require("..//lib/main.js");
```
zorkow: ditto

test/issue215.js
```diff
@@ -1,5 +1,5 @@
 var tape = require('tape');
-var mjAPI = require("..//lib/mj-single.js");
+var mjAPI = require("..//lib/main.js");
```
zorkow: ditto

test/svg-metadata.js
```diff
@@ -1,5 +1,5 @@
 var tape = require('tape');
-var mjAPI = require("..//lib/mj-single.js");
+var mjAPI = require("..//lib/main.js");
```
zorkow: ditto

pkra:
Thanks @zorkow I'll fix them in the morning.

pkra:
Ok, I've fixed the paths. 

@dpvc are you ok with the name change to `main.js`?

dpvc:
> are you ok with the name change to `main.js`?

Yes, that's fine.  I might have gone with `mathjax.js`, but since I never commented on that, I can't really complain.

pkra:
> I might have gone with mathjax.js.

I'm ok with that.

pkra:
@zorkow are you ok with naming it `mathjax.js`?

pkra:
Merging as is -- as per F2F. 

@dpvc I didn't have time to bring up https://github.com/mathjax/MathJax-node/pull/247#issuecomment-228728321. Could you check that out some time?

dpvc:
@pkra, I only looked at it briefly, and do have comments, but haven't had the chance to write them up.  I'll do it as soon as I can.

pkra:
Thanks, @dpvc. No rush. Just wanted to check you saw it.

dpvc:
I took another look at your code, and am not sure I fully understand what you are after.  Can you say more about what you are trying to accomplish?  We can use the gist page if you want (but let me know so that I can look -- I don't think I get notification from it until I post there).  

One thing I noted was that you don't load an output renderer, so MathJax will certainly crash on that.  But, of course, you can't use the output jax as they stand, because they all use the DOM to do some measurements.  That is why `mj-page` (and `mj-single`) have to override those.  So I'm a bit confused by what you are trying to do.  It looks to me like you are going to have to move the `AuthorInit` from `mj-page`, or most of it, into your new tool.  So I'm not sure how that reduces maintenance (it just seems to put things into the hands of the tool author, where we will have to answer more questions about it, and where they will make modifications that we will have to try to understand when they ask about it).  So I don't know what's up with this.

If you are trying to get tex2jax and the other preprocessors to locate the math expressions in the page, you may be able to do that, but this isn't quite the right way to do it.  I can give you some pointers if that is what you want.  My expectation was that you would be trying to find the math in the document and then pass it to `mj-single`, but you seem to be trying to run MathJax on the page directly.  That was what `mj-page` does, so that would require essentially all of `mj-page` (after all, that is what it was for).   Can you help me understand better what you are trying to do?

Also, you say "Merging as is -- as per F2F", but I don't recall having a face-to-face about this.  In fact, this PR wasn't made until the day before I left for vacation, and the only F2F we've had since then was Wednesday's hangout.  I don't recall discussing it then.  Have I forgotten something?

pkra:
Thanks, Davide.

> If you are trying to get tex2jax and the other preprocessors to locate the math expressions in the page, you may be able to do that, but this isn't quite the right way to do it. 

That is what I'm after. Just pre-process the page to find some math.

> Also, you say "Merging as is -- as per F2F", but I don't recall having a face-to-face about this. In fact, this PR wasn't made until the day before I left for vacation, and the only F2F we've had since then was Wednesday's hangout. I don't recall discussing it then. Have I forgotten something?

Sorry that we had a miscommunication here! On the one hand, I thought given @zorkow's review the merge as such had been cleared. What had remained, in my mind, was the naming issue, which we discussed at this week's F2F. I thought we had agreed there that we won't switch to `mathjax.js` after all. 

Do you want to revert for the time being?

pkra:
FYI I've updated https://gist.github.com/pkra/1ae75a4bb86fbc3364c8d85675030223 to use a modified tex2jax directly.

PR #249, opened 2016-06-28
Adds string conversion for xml output of semantic tree. Fixes issue #…

zorkow:
…248.


pkra:
**Note**. This will cause a merge conflict with #247 (where mj-single is renamed to main)

Should this rather be handled as part of #207?

zorkow:
Makes sense. However, it is just a bug fix, which we might want to add earlier.

pkra:
> Makes sense. However, it is just a bug fix, which we might want to add earlier.

No objections. Let's just wait until #247.

Issue #219, opened 2016-04-20
return an object not a serialization


pkra:
To separate this out from #205. 

As a developer, I want to post-process the results from mathjax-node which is easier to do with a DOM-like object rather than a string.

The problem is to find a reasonably universal, DOM-like object format in nodejs-land. Right now we're stuck with jsdom anyway which, I think, means parse5.

Issue #208, opened 2016-04-11
[discussion] re-organizing the examples in `/bin`


pkra:
As per F2F, the examples in `/bin` have been growing and becoming a burden to maintain.

A couple of ideas to move forward:
- turn them it into a separate repository and node module (or several modules), hosted but not officially maintained by the MathJax team (cf. the MathJax third-party extension repository). This could also open ta path to a refactoring, e.g., into one unified tool (though that might increase the maintenance load going forward).
- replace them some other way, e.g., as a collection of recipes like [gulp's](https://github.com/gulpjs/gulp/tree/master/docs/recipes)

Any other suggestions?

pkra:
I've created an example repice in #119; I'm growing to the idea.

dpvc:
I think that making them into a single tool is probably the right approach.  I don't have strong feelings about whether that should be in a separate repository or not.

I don't see a collection of recipes as being any less burdensome than the current collection of tools.

pkra:
> I think that making them into a single tool is probably the right approach. 

Fair enough. I just don't think it's a core product.

> I don't see a collection of recipes as being any less burdensome than the current collection of tools.

Right (though I would drop most of them); I do think that having them in `./bin` gives people the wrong idea about their status (as we've seen in bug reports). Examples in a documentation have a different status; one that fits better, I think.

pkra:
How about this as a compromise: rename the folder "examples" instead of "bin"?

pkra:
https://github.com/mathjax/MathJax-node/issues/119#issuecomment-209324821 and https://github.com/mathjax/MathJax-node/issues/205#issuecomment-221511243 relate to this problem. They make good examples but no more.

michael-okeefe:
If it helps, as a somewhat anecdotal example, I came to this repository after much Google searching and have been using the scripts in the bin/ directory as a pathway to pre-render MathJax for a static website.

I'm not sure how many other people are in the same boat but I barely know node.js and found these ready-to-run scripts to be extremely useful. One problem I'm running into is that the default script for page2svg runs rather slow although that's a separate issue and one I'm willing to put up with.

If my use-case (people looking for a pathway to command-line access to mathjax pre-rendering who don't know node.js well) is not a complete outlier, I would recommend and would personally love to see more effort put into the bin/ scripts to make sure they are robust, performant, and well-documented. Seems like you have a good start already.

[...]

PR #250, opened 2016-06-29
Removes sre from mathjax node.

zorkow:
The PR removes SRE from MathJax-node core as discussed in issue #207. Here are the main changes and/or points to discuss:
- All references to speech are taken out of main.js
- A new speech.js file offeres some helper functionality. However, I could see the main method that injects speech into a DOM node move into sre api.
- As examples all the bin/tex2\* have been adapted to the new code.
- I have introduced new htmlElement, svgElement, mmlElement fields to return the actual DOM elements in the data object. Note, that the mmlElement has to be explicitly created.
- SVG now comes with title and desc element by default. The latter contains the original input code. Not sure if that makes sense in the MathML case.

### DIFF
--- a/bin/tex2html
+++ b/bin/tex2html
@@ -24,6 +24,7 @@
  */
 
 var mjAPI = require("../lib/main.js");
+var speech = require('../lib/speech.js');                    //  Speech integraton helper functions.
 
 var argv = require("yargs")
   .demand(1).strict()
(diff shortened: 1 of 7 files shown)


zorkow:
Two sre related tests fail.

zorkow:
To avoid misunderstandings following a hangout this morning:

The speech.js is not meant to stay, but should be dissolved eventually, moving functionality either into sre or an npm package of its own. The PR is more about discussing if we are happy with the new main.js and how speech can be integrated in the example scripts.

pkra:
Just some feedback.

* I would have approached the the clean-up of SRE differently. Basically, I'd leave it as is but instead of getting speech text from SRE, I'd use a) an alt-text input (from an option, from mathml alttext via MathJax), b) TeX or asciimath source, c) some generic fallback text to fill in the text. That way, it not only works as a post-processor but also as a pre-processor (e.g., when doing semantic enrichment, I'd like the enhaned MathML and the speech so why pass it back later).
* the fix for #219 looks good to me but I'd prefer to do it separately
* speech.js looks good to me; it should be a separate package and repo
* the changes to the examples in `bin` only remind me that #208 needs some agreement

pkra:
Closing in favor of #273

PR #252, opened 2016-07-11
[main] add option for returning dom objects

### LINKED ISSUE #219: return an object not a serialization

pkra:
Fixes #219.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -54,9 +54,12 @@ var defaults = {
   xmlns: "mml",                   // the namespace to use for MathML
 
   html: false,                    // return HTML output?
+  htmlElement: false,             // DOM element for HTML output?
   css: false,                     // return CSS for HTML output?
   mml: false,                     // return mml output?
+  mmlElement: false,              // DOM element for MML output?
   svg: false,                     // return svg output?
+  svgElement: false,              // DOM element for SVG output?
 
   speakText: false,               // add spoken annotations to svg output?
   speakRuleset: "mathspeak",      // set speech ruleset (default (chromevox rules), mathspeak)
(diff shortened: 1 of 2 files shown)


pkra:
A slightly different approach to @zorkow's in #250. 

PTAL.

dpvc:
Note that the `*Element` items only work if the corresponding non-element forms are also selected.  So you can't get *just* the SVG element, for example, without getting the SVG text as well.  Is that intended?

Does this mean that `svg` should be renamed `svgText`, etc.?

Also, would you consider `svgNode` rather than `svgElement`, since it is shorter?

pkra:
> Is that intended?

Not really. Just a result of being lazy. I will work that in.

> Does this mean that svg should be renamed svgText, etc.?

I'd be happy to find some better names. I just copied those from @zorkow's PR. I could also imagine to have `svg.text` and `svg.node` but I'm not very good at naming things :smile: 

> Also, would you consider svgNode rather than svgElement, since it is shorter?

That sounds good to me.

pkra:
I've updated the branch to address most comments.

What I have not addressed yet is @dpvc's question whether `svg` should become `svgText` . I'm not sure whether that's necessary -- but it's a 3/10 for me so I'm ok with changing it if we decide that. Any thoughts?

pkra:
As per F2F, ready to go.

PR #258, opened 2016-08-10
Fixed bug with equation numbers not resetting to 0

CemraJC:
If `typeset()` was called multiple times *without* a call to `start()` in between, the equation numbering does not reset to 0. Unfortunately, always calling `start()` causes undefined behaviour.

In my tests for [gulp-mathjax-node](https://github.com/cemrajc/gulp-mathjax-node), I found this was an issue with the `mj-page.js` file.

Admittedly, this is a *very* hard-to-spot error :smile:

### DIFF
--- a/lib/mj-page.js
+++ b/lib/mj-page.js
@@ -641,7 +641,7 @@ function ConfigureTypeset() {
   //  Reset the MathJax counters for things
   //
   SVG.resetGlyphs(true);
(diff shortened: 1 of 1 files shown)


pkra:
Thanks for spotting this! I'm closing this for two technical reasons. One the one hand, PRs need to be made to the `develop` branch. On the other hand, `develop` no longer has mj-page in it; see #206 for more information.

PR #262, opened 2016-09-26
[main] getSVG should increment state.ID

### LINKED ISSUE #260: state.ID does not increment
This is because the state is update in `TypesetDone` before `GetSVG` runs (to increment it).

Relates to #207.

pkra:
* move state handling from TypesetDone to ReturnResult
Fixes #260


pkra:
Thanks, @dpvc!

PR #264, opened 2016-09-28
Change and fix README

kasperpeulen:
I think the commits speak for itself :)


pkra:
Thanks. I'm afraid PRs should be made to `develop` -- but that has changed enough to make this change obsolete (though we are still discussing whether mj-page should make a come back in this repository or be its own module).

Issue #248, opened 2016-06-28
XML output for semantic tree is broken


zorkow:
Missing string conversion in GetSemantic in mj-single

zorkow:
Should be fixed after file structure refactoring is merged.

pkra:
I can't find a branch with this commit so I've made a new one.

PR #269, opened 2016-10-11
[main] GetSemantic: add string conversion

pkra:
* add string conversion for xml output of semantic tree. Fixes issue #248
* add test for GetSemantic


pkra:
As per F2F, ready to go.

PR #270, opened 2016-10-11
[main] pass input data to result object

### LINKED ISSUE #239: [mj-single] passing data through to the result object
In the context of #205, #206, #207 and other work, I've been wondering about passing data through to the result object. I also wondered why the results do not include the input information (source, format etc)

Anyway, all I could come up with was abusing `state` but that seems like a limited and generally bad approach. Is there a better way? Is this just a terrible idea?

pkra:
Resolves #239


pkra:
As per F2F, ready to go.

PR #271, opened 2016-10-27
Prevent the color extension from auto-loading

### LINKED ISSUE #175: Prevent the `color` extension from auto-loading
From https://github.com/mathjax/MathJax-node/issues/143#issuecomment-168243285.

> You are right to point out the `autoload-all` extension, and I was going to say that because `\color` is already defined in MathJax's TeX input jax by default, that `autoload-all` would not redefine it, but I checked the code and I am wrong; `\color` is set to autoload the color extension.  I'm not sure that is the desired behavior for MathJax-node, since that means it will operate differently from default MathJax in the browser.

pkra:
Fixes #175

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -427,6 +427,20 @@ function ConfigureMathJax() {
       });
 
       //
+      //  Reset the color extension after `autoload-all`
+      //
+      if (MathJax.AuthorConfig.extensions.indexOf("TeX/color.js") == -1) {
+              MathJax.Hub.Register.StartupHook("TeX autoload-all Ready",function () {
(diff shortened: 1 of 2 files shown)


dpvc:
I'm wondering if a test that checks the result of something like `\color{red}{x}x` might be worth adding, since with the original `autoload-all`, this would make both x's red, while with the modification, it should only make the first one red.

pkra:
> I'm wondering if a test that checks the result of something like \color{red}{x}x might be worth adding, 

Will do.

pkra:
New test added.

pkra:
ping @dpvc

pkra:
Thanks!

Issue #266, opened 2016-10-11
compatibility with latest jsdom


pkra:
It looks like our patching fails with jsdom 9.6.0.

pkra:
@dpvc could you take a look?

dpvc:
I have made a pull request to resolve the issue.  The patch should work with both jsdom version 9 and version 7.

pkra:
Thanks!

PR #272, opened 2016-10-29
Fix jsdom patch to work with version 9.  #266

dpvc:
Fix jsdom patch to work with version 9 as well as version 7, and set user agent since the one in version 9 now makes MathJax think it is Safari.  Resolves issue #266.


pkra:
Could you update package.json with the jsdom version you've tested against? (Maybe we should make it a range such as `>=7.0 <=9.8`, cf https://docs.npmjs.com/files/package.json#dependencies?)

dpvc:
OK, done.

pkra:
Thanks!

PR #273, opened 2016-11-01
remove speech-rule-engine integration

### LINKED ISSUE #207: move SRE integration out of mj-node

pkra:
Resolves #207


pkra:
Ready for review.

pkra:
@dpvc @zorkow I did a [poll on twitter](https://twitter.com/pkrautz/status/793752948014710784) (some people responded in tweets, but those were all against "prefixes"); not exactly scientific but I think we should drop the equation title. What do you think?

(FWIW I have been thinking about proposing a redefinition of role=math, and prefixes would be one item that AT could provide an opt-in for.)

pkra:
Also, https://github.com/pkra/mathjax-node-sre is now actually working.

pkra:
> I think we should drop the equation title. What do you think?

I've updated the commit.

lib/main.js
```diff
@@ -549,30 +540,18 @@ function GetSpeech(result) {
...
   if (!data.html && !data.htmlNode) return;
   var jax = MathJax.Hub.getAllJax()[0]; if (!jax) return;
   var script = jax.SourceElement(), html = script.previousSibling;
 
-  if (data.speakText) {
-    var label = html.querySelector("[aria-label]");
-    if (label) label.removeAttribute("aria-label");
+  // add dummy speech text if there isn't one
+  for (var i=0, m=html.childNodes.length; i < m; i++)
+    html.childNodes[i].setAttribute("aria-hidden",true);
```
zorkow: See remark below.

lib/main.js
```diff
@@ -608,17 +587,16 @@ function GetSVG(result) {
...
-    svg.setAttribute("aria-labelledby",id+"-Title "+id+"-Desc");
-    for (var i=0, m=svg.childNodes.length; i < m; i++)
-      svg.childNodes[i].setAttribute("aria-hidden",true);
-    var node = MathJax.HTML.Element("desc",{id:id+"-Desc"},[result.speakText]);
-    svg.insertBefore(node,svg.firstChild);
-    node = MathJax.HTML.Element("title",{id:id+"-Title"},["Equation"]);
-    svg.insertBefore(node,svg.firstChild);
-  }
+  for (var i=0, m=svg.childNodes.length; i < m; i++)
+    svg.childNodes[i].setAttribute("aria-hidden",true);
```
zorkow: In future coding style this should become a utility function (e.g., ariaHideChildren(node)), together as in lines 551/552 above.

lib/main.js
```diff
@@ -608,17 +587,16 @@ function GetSVG(result) {
...
+  // Note: if aria-label exists, getSpeech preserved it in speakText
+  // remove aria-label since labelled-by desc is preferred
+  svg.removeAttribute("aria-label");
```
zorkow: Do we just throw away the original aria label, if there is one?

test/issue207.js
```diff
@@ -0,0 +1,52 @@
...
+    var element = document.querySelector('.MJXc-display');
+    var result = element.getAttribute('aria-label');
+    t.equal(result, input1, 'HTML output contains dummy speechText');
+    var document = jsdom(data.mml).defaultView.document;
+    var element = document.querySelector('math');
+    var result = element.getAttribute('alttext');
+    t.equal(result, input1, 'MathML output contains dummy speechText');
+    var document = jsdom(data.svg).defaultView.document;
+    var desc = document.querySelector('desc');
+    t.equal(result, desc.innerHTML, 'SVG output contains dummy speechText');
```
zorkow: Refactor into function with code in the second unit test.

test/issue207.js
```diff
@@ -0,0 +1,52 @@
...
+    var result = element.getAttribute('aria-label');
+    t.equal(result, expected, 'HTML output from MathML contains original speechText');
+    var document = jsdom(data.mml).defaultView.document;
+    var element = document.querySelector('math');
+    var result = element.getAttribute('alttext');
+    t.equal(result, expected, 'MathML output from MathML contains original dummy speechText');
+    var document = jsdom(data.svg).defaultView.document;
+    var desc = document.querySelector('desc');
+    t.equal(expected, desc.innerHTML, 'SVG output from MathML contains original dummy speechText');
+  });
```
zorkow: Refactor, see remark above.

test/issue207.js
```diff
@@ -0,0 +1,52 @@
...
+    mml: true
+  }, function(data) {
+    var document = jsdom(data.html).defaultView.document;
+    var element = document.querySelector('.MJXc-display');
+    var result = element.getAttribute('aria-label');
+    t.equal(result, expected, 'HTML output from MathML contains original speechText');
+    var document = jsdom(data.mml).defaultView.document;
+    var element = document.querySelector('math');
+    var result = element.getAttribute('alttext');
+    t.equal(result, expected, 'MathML output from MathML contains original dummy speechText');
```
zorkow: Call parameters actual and expected. 
(Convention dictacted that expected comes before actual, but that's just silly...)

lib/main.js
```diff
@@ -532,15 +523,15 @@ function GetMML(result) {
...
 function GetSpeech(result) {
-  if (!data.speakText) return;
-  speech.setupEngine({semantics: true, domain: data.speakRuleset, style: data.speakStyle});
-  result.speakText = speech.toSpeech(result.mml);
+  result.speakText = "Equation";
+  if (data.format !== "MathML") result.speakText = data.math;
   if (!data.mml) return;
   var jax = MathJax.Hub.getAllJax()[0];
-  jax.root.alttext = result.speakText;
+  if (jax.root.alttext) result.speakText = jax.root.alttext;
```
pkra: I don't understand. The next line sets the `alttext` if it didn't exist.

pkra: Since I'm not sure how the reviews are pushed out in notifications: I just drastically edited my earlier comment.

lib/main.js
pkra: Yes because your recommendation had been to use `desc` instead of an aria-label for SVGs.

The aria-label occurs only via core mathjax: if MathML input has an `alttext`, core mathjax adds that as an aria-label. But in that case we used the `alttext` value for `speakText` above and thus the `desc` will be the same as the label. Moving the change upstream would eliminate this part.

test/issue207.js
pkra: Hm. I don't see how to refactor this without making the test worse. I don't want to give up on having differentiating messages from the tests. I also don't want to create some complicated function that generates fitting tests and messages dynamically because I worry that this will make the test less readable. 

Do you have any suggestions?

test/issue207.js
pkra: > Call parameters actual and expected.

OK.

test/issue207.js
pkra: See above.

lib/main.js
zorkow: What I meant was if alttext is "" rather than non-existent, then it will be overwritten by speakText. Could the case happen that alttext is ever deliberately left empty? If yes, then the test should be refined to check for undefinedness, if no, it's fine.
[This is just a sanity-check. Sometimes careless treatment of semi-boolean values, i.e., "", '', 0, can lead to unintended effects and errors. I ran often enough into that problem, e.g,. when iterating over lists of integers.]

lib/main.js
zorkow: sgtm.

lib/main.js
pkra: Thanks for clarifying! I had misunderstood and had not thought of this imprecision. I think overwriting falsy values is ok but I'm happy to change it.

test/issue207.js
zorkow: In addition you could refactor the actual calls to typeset into a single function, so the actual calls to the test function will be of the form:
`callMjSpeechTest(format, math, expected);`
e.g., `callMjSpeechTest('TeX', 'x', 'x');`
which will make the code and the principle of the tests a lot more readable.

pkra: Thanks for clarifying. As I wrote earlier, the problem I saw with this kind of refactor is that the tap messages are losing important information. But I have an idea for working around that.

zorkow:
Certainly weird. It seems to duplicate comments.

pkra:
I mostly dislike that it opens a new review when I'm just trying to respond to your comments. Especially since I'd have assumed that I can't review my own PR.

zorkow:
Well, I sometimes review my own PRs in SRE, but remember github complaining, that I can't open a review on my own PR. But apparently you can review someone's review on your PR. Is that a meta-review???

Anyway, I would have thought that it clearly groups replies chronologically, if that makes sense.

pkra:
@zorkow I've refactored the test.

lib/main.js
```diff
@@ -60,14 +59,6 @@ var defaults = {
   mmlNode: false,              // DOM element for MML output?
   svg: false,                     // return svg output?
   svgNode: false,              // DOM element for SVG output?
-
-  speakText: false,               // add spoken annotations to svg output?
```
dpvc: Your current implementation always returns `speakText` in the `results` object, and always adds some speech text to the SVG, HTML, or MML output.  I'd prefer to retain the option of adding speech text (though it could be true by default), and only add it and return it when requested.

lib/main.js
```diff
@@ -532,15 +523,15 @@ function GetMML(result) {
...
 //
-//  Creates the speech string and updates the MathML to include it, if needed
+//  Creates dummy speech string and updates the MathML to include it, if needed
 //
 function GetSpeech(result) {
-  if (!data.speakText) return;
-  speech.setupEngine({semantics: true, domain: data.speakRuleset, style: data.speakStyle});
-  result.speakText = speech.toSpeech(result.mml);
+  result.speakText = "Equation";
+  if (data.format !== "MathML") result.speakText = data.math;
```
dpvc: Note that it is possible for this to make the `speakText` be blank (if `data.math` is blank, e.g., from `\(\)` and input is TeX).

lib/main.js
```diff
@@ -608,17 +587,16 @@ function GetSVG(result) {
...
-    svg.insertBefore(node,svg.firstChild);
-    node = MathJax.HTML.Element("title",{id:id+"-Title"},["Equation"]);
-    svg.insertBefore(node,svg.firstChild);
-  }
+  for (var i=0, m=svg.childNodes.length; i < m; i++)
+    svg.childNodes[i].setAttribute("aria-hidden",true);
+  // Note: if aria-label exists, getSpeech preserved it in speakText
+  // remove aria-label since labelled-by desc is preferred
+  svg.removeAttribute("aria-label");
+  ID++; var id = "MathJax-SVG-"+ID;
```
dpvc: Why not just add `'-Desc'` to the id at this point rather than adding it twice below (now that you no longer need `id+'-Title'`)?

dpvc:
Here are a few other comments that I couldn't add as line comments, since they are not about lines in the change file.

* The blank line at [600](https://github.com/mathjax/MathJax-node/blob/444451543ab51b76f23c709f630d3c355c6ea6a6/lib/main.js#L600) is not needed.  (There are some other spacing issues that Volker may complain about as well...)

* [Lines 740 to 742](https://github.com/mathjax/MathJax-node/blob/444451543ab51b76f23c709f630d3c355c6ea6a6/lib/main.js#740-742) are  no longer needed, since the `results.mml` is only created when actually needed (it used to be created for speech creation even when not requested by `data.mml`, and had to be cleaned up here).  If you end up going with optional speech string creation, as I requested above, then you might use similar code to this to remove `results.speakText` when not needed.

* The speech options need to be removed from the command-line tools.

I looked over the file as a whole, and found some issues that don't come directly from this pull request, but should be handled at some point anyway:

* At [line 606](https://github.com/mathjax/MathJax-node/blob/444451543ab51b76f23c709f630d3c355c6ea6a6/lib/main.js#L606-L613), you no longer need a separate variable (it used to be used to be needed for writing the external files, but that was removed earlier).

* At [line 573](https://github.com/mathjax/MathJax-node/blob/444451543ab51b76f23c709f630d3c355c6ea6a6/lib/main.js#L573) and [line 615](https://github.com/mathjax/MathJax-node/blob/444451543ab51b76f23c709f630d3c355c6ea6a6/lib/main.js#L615), the nodes being returned should be removed from the DOM before being sent back in the `results` object (otherwise the caller then has access to the internal jsdom document).  Alternatively, they could be cloned rather than disrupting the DOM.

* At [line 718](https://github.com/mathjax/MathJax-node/blob/444451543ab51b76f23c709f630d3c355c6ea6a6/lib/main.js#L718), `MathJax.OutputJax.CommonHTML` can be replaced by `CHTML`, since that is defined earlier.

* Similarly, at [line 750](https://github.com/mathjax/MathJax-node/blob/444451543ab51b76f23c709f630d3c355c6ea6a6/lib/main.js#L750), `MathJax.ElementJax.mml` can be replaced by `MML`.

* In [lines 616 through 618](https://github.com/mathjax/MathJax-node/blob/444451543ab51b76f23c709f630d3c355c6ea6a6/lib/main.js#L616-618), should these be `svgWidth`, `svgHeight`, and `svgStyles`, since they are specific to the SVG output?

These are what I see at this point.  Hope that helps.

lib/main.js
pkra: OK

lib/main.js
pkra: Right. Cf. also Volker's comment below.

lib/main.js
pkra: No reason. Just copy and paste from the original (where a title was also present and thus you had split this up).

pkra:
> The speech options need to be removed from the command-line tools.

Since you suggested the switch should stay, they should stay as well.

> I looked over the file as a whole, and found some issues that don't come directly from this pull request, but should be handled at some point anyway:

I suggest we file those separately.

pkra:
I hope I've addressed all review comments.

pkra:
Sorry. I broke the tests (again).

pkra:
Fixed the tests.

lib/main.js
dpvc: Thanks.  Note, however, that the comment says it adds to svg, but it really adds to all output formats.

dpvc:
> >The speech options need to be removed from the command-line tools.
>
> Since you suggested the switch should stay, they should stay as well.

There were several options to control SRE that are no longer needed and should be removed from them.  Also, now that the default for `speakText` is `true`, you may want the command line tools to default that way as well, which they currently don't.

dpvc:
Also, you will want to remove the `*2stree` command-line tools, since those functions have been removed.

lib/main.js
```diff
@@ -608,17 +591,17 @@ function GetSVG(result) {
...
-    node = MathJax.HTML.Element("title",{id:id+"-Title"},["Equation"]);
-    svg.insertBefore(node,svg.firstChild);
+    // Note: if aria-label exists, getSpeech preserved it in speakText
+    // remove aria-label since labelled-by desc is preferred
+    svg.removeAttribute("aria-label");
   }
+  ID++; var id = "MathJax-SVG-"+ID+"-Desc";
+  svg.setAttribute("aria-labelledby",id);
+  var node = MathJax.HTML.Element("desc",{id:id},[result.speakText]);
+  svg.insertBefore(node,svg.firstChild);
```
dpvc: Shouldn't these four lines (601-604) be inside the test for `data.speakText` as well?

pkra: Right.

pkra:
> There were several options to control SRE that are no longer needed and should be removed from them. 

Right. But cf #208, I still think they should all live in an entirely separate package anyway.

pkra:
> Also, now that the default for speakText is true, you may want the command line tools to default that way as well, which they currently don't.

Looks like they do?

pkra:
PTAL.

pkra:
> Looks like they do?

Ah, I was wrong about that. Confusing the type with the default. Will add a default.

pkra:
PTAL.

pkra:
I've been doing [a bit of testing today](https://s.codepen.io/pkra/debug/RodYOJ) and realized that we should change the SVG approach: a `title` tag has a better chance of actually producing a result on AT.

pkra:
PTAL.

dpvc:
I'm OK with it at this point.  @zorkow needs to accept your changes to complete his review.  Then we can merge.

test/issue207.js
zorkow: Don't redeclare variables in the same scope. Otherwise, the interpreted does a lot of hoisting, which is unnecessary.

zorkow:
Get rid of the superfluous "var" declarations. Then it's OK.

pkra:
> Get rid of the superfluous "var" declarations. Then it's OK.

Uhm. Which commit were you seeing these on? I have trouble spotting them :-(

pkra:
> Get rid of the superfluous "var" declarations. Then it's OK.

Cleared up F2F.

PR #276, opened 2016-11-10
Add missing pxPerInch value.  #275

### LINKED ISSUE #275: [SVG output] mspace with cm units behaving badly
E.g., `E_iE_j=E_jE_i,\quad F_iF_j=F_jF_i,\quad \hspace{2.1cm} {\mathrm{if}}\ |i-j|>1` renders fine in the browser but in mathjax-node the content beyond the mspace is move to the beginning of the line, overlapping with the other content.

dpvc:
This defines the `pxPerInch` value needed for absolute units like `cm`, `mm`, `in`, etc. that is usually computed by measuring DOM elements.  Resolves #275.

### DIFF
--- a/test/issue276.js
+++ b/test/issue276.js
@@ -1,0 +1,20 @@
+var tape = require('tape');
+var mjAPI = require("../lib/main.js");
+var jsdom = require('jsdom').jsdom;
+
+tape('SVG output: physical units', function(t) {
+  t.plan(1);
+  mjAPI.start();
+  var mml = '<math><mspace width="1cm"></mspace></math>';
+
+  mjAPI.typeset({
+    math: mml,
+    format: "MathML",
+    svg: true
+  }, function(data) {
+    var document = jsdom(data.svg);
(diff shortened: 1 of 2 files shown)


pkra:
I'll add a test.

pkra:
@dpvc PTAL.

test/issue276.js
```diff
@@ -0,0 +1,21 @@
...
+
+  mjAPI.typeset({
+    math: mml,
+    format: "MathML",
+    svg: true
+  }, function(data) {
+    var document = jsdom(data.svg);
+    var doc = document.defaultView.document;
+    var width =  doc.querySelector('svg').getAttribute('width');
+    console.log(width);
```
dpvc: Is this left-over debugging, or do you really mean to print the width?

pkra: Ah. Thanks for catching that.

pkra:
I updated the test.

dpvc:
OK, then it looks good.  Merge when ready.

PR #284, opened 2016-12-29
Don't produce mml when mmlNode is requested if mml wasn't requested. #277

### LINKED ISSUE #277: should mmlNode:true force mml:true?
`mmlNode` requires `mml` but not vice versa (whereas it's the reverse for the other outputs).

dpvc:
Don't produce `mml` when `mmlNode` is produced (unless `mml` is also requested).  Fixes a problem with PR #282, and resolves #277.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -537,12 +537,13 @@ function GetMML(result) {
     }
   }
   try {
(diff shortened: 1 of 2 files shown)


pkra:
Ah, now I understand what you meant!

PR #289, opened 2017-02-08
[main.js] add missing checks for htmlNode

### LINKED ISSUE #288: Issues when generating SVG and HTML simultaneously
On the `develop` branch:

* When setting `htmlNode` and `svgNode`, `result.htmlNode` is populated with SVG outuput (including the wrapping span)

```javascript
 mjAPI = require("../lib/main.js");
  mjAPI.typeset({
        math: tex,
        format: "TeX",
        htmlNode: true,
        svgNode: true
    }, function(data) {
      console.log(data.htmlNode.outerHTML); 
     // => <div class="MathJax_SVG_Display" style="text-align: center;" aria-label="x"><svg ... etc.
    });
```

* When using `html` and `svgNode`, an error is thrown

```javascript
 mjAPI = require("../lib/main.js");

 mjAPI.typeset({
     math: 'x',
     format: "TeX",
     html: true,
     svgNode: true
 }, function(data) {
     console.log(data.html);
 });
// => Error: TypeError: Cannot read property 'setAttribute' of undefined
//    at GetSVG (/home/peter/code/mathjax/node/lib/main.js:605:6)
```

pkra:
Fixes #288

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -684,7 +684,7 @@ function StartQueue() {
   //
   GetState(data.state);
 
-  var renderer = (data.html ? "CommonHTML" : "SVG");
+  var renderer = ( (data.html || data.htmlNode) ? "CommonHTML" : "SVG");
 
   //
   //  Set up a timeout timer to restart MathJax if it runs too long,
@@ -784,7 +784,7 @@ function SetRenderer(renderer) {
 }
 
 function RerenderSVG(result) {
-  if (data.html && data.svg) {
+  if ((data.html || data.htmlNode ) && (data.svg || data.svgNode)) {
     timer = setTimeout(RestartMathJax,data.timeout);
     var queue = MathJax.Callback.Queue(), $$ = window.Array;
     return queue.Push(
--- a/test/issue288.js
+++ b/test/issue288.js
@@ -1,0 +1,26 @@
+var tape = require('tape');
+var mjAPI = require("../lib/main.js");
+
+tape('HTML output when requesting both SVG and HTML', function(t) {
+    t.plan(2);
+    mjAPI.start();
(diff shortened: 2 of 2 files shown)


dpvc:
This looks OK, but I think that there may need to be one more change in order to handle requests for `css` when there is no request for `html` or `htmlNode`.  You may want to move [line 594](https://github.com/mathjax/MathJax-node/blob/f1946b2aee3d3eab96fb5aa13e7e8f7f116fa9a2/lib/main.js#L594) to the top of that function, right before [line 566](https://github.com/mathjax/MathJax-node/blob/f1946b2aee3d3eab96fb5aa13e7e8f7f116fa9a2/lib/main.js#L566).

pkra:
Ah! And I'm guessing https://github.com/mathjax/MathJax-node/blob/f1946b2aee3d3eab96fb5aa13e7e8f7f116fa9a2/lib/main.js#L687 needs to check for `css` as well so that typesetting happens with CHTML and the styles get generated, right?

dpvc:
No, the styles are obtained just by loading the output jax, and that always happens.  So you don't need to typeset using CHTML to get that.  (I haven't tested that, but I'm pretty sure that's true.)

pkra:
> I haven't tested that, but I'm pretty sure that's true.

I've tested it and it fails for me, see #292. I think we should merge this and deal with the other issue separately.

dpvc:
Since the fix for `data.css` is in the same line (687) as this change, I'd say doing as part of this PR would be reasonable.

pkra:
In the interest of concise commits, can I go ahead and merge?

dpvc:
OK, sure.

PR #291, opened 2017-02-08
[main.js] getHTML: add aria-label to correct childnode

pkra:
Resolves #289

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -570,8 +570,7 @@ function GetHTML(result) {
   // add speech text if there isn't one
   if (data.speakText){
     var labelTarget = html.querySelector('.mjx-math');
-    for (var i=0, m=labelTarget.childNodes.length; i < m; i++)
-      labelTarget.childNodes[i].setAttribute("aria-hidden",true);
+    for (child of labelTarget.childNodes)     child.setAttribute("aria-hidden",true);
     if (!labelTarget.getAttribute("aria-label")){
     labelTarget.setAttribute("aria-label",result.speakText);
     }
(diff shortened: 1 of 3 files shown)


pkra:
Tests fixed.

lib/main.js
```diff
@@ -569,9 +569,12 @@ function GetHTML(result) {
 
   // add speech text if there isn't one
   if (data.speakText){
-    for (var i=0, m=html.childNodes.length; i < m; i++)
-      html.childNodes[i].setAttribute("aria-hidden",true);
-    if (!html.getAttribute("aria-label")) html.setAttribute("aria-label",result.speakText);
+    var labelTarget = html.querySelector('.mjx-math');
+    for (var i=0, m=labelTarget.childNodes.length; i < m; i++)
```
zorkow: for (var i = 0, child; child = childNode[i]; i++) {
...
}

pkra:
Question: why was the label added twice? shouldn't speakText stop when there's alttext from mathml?

pkra:
We do add speakText twice if mml with alttext is the input, which technically isn't necessary. I'm ok with it as is. 

A better solution would be nice, i.e., updating the internal format and leaving it to mathjax. But that would require some refactoring of the chain of events, e.g., splitting the mathjax processing so we can update the internal format; that seems out of scope here.

lib/main.js
pkra: PTAL.

pkra:
@zorkow ptal

pkra:
Ok per F2F

Issue #285, opened 2016-12-29
More thorough tests


dpvc:
When making the changes for PR #284, I had to change the test file for it, and that got me thinking about the tests for mathjax-node.  Currently, the tests like this one only check that a particular field is present in the results, but a more thorough test would be to check that exactly the expected fields are present.  So if you asked for `mmlNode`, one would not just check for `results.mmlNode`, but also that no other values were present in `results`.  What do you think?

I suppose what we really need is a comprehensive test suite that tests the actual features and their expected results.  All we have now are tests for specific issues and their fixes.

pkra:
On a related note, stratifying test titles and filenames is something I've been itching to do.

PR #296, opened 2017-02-16
[main] enable css output without html output

### LINKED ISSUE #292: No css output without html or htmlNode
Extracting this from https://github.com/mathjax/MathJax-node/pull/289#issuecomment-279821382

>  [...] to handle requests for `css` when there is no request for `html` or `htmlNode`.  You may want to move [line 594](https://github.com/mathjax/MathJax-node/blob/f1946b2aee3d3eab96fb5aa13e7e8f7f116fa9a2/lib/main.js#L594) to the top of that function, right before [line 566](https://github.com/mathjax/MathJax-node/blob/f1946b2aee3d3eab96fb5aa13e7e8f7f116fa9a2/lib/main.js#L566).

 If I make the change, the following test still fails.

```javascript
 var tape = require('tape');
var mjAPI = require("../lib/main.js");

tape('CSS output', function(t) {
    t.plan(1);
    mjAPI.start();
    var tex = 'x';
    mjAPI.typeset({
        math: tex,
        format: "TeX",
        css: true
    }, function(data) {
        t.ok(data.css, 'css output while no other output');
    });
});
```
Digging a little bit, `CHTMLSTYLES` is undefined at that point. Maybe it's a restarting issue?

pkra:
Resolves #292

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -563,6 +563,7 @@ function GetSpeech(result) {
 //  Create HTML and CSS output, if requested
 //
 function GetHTML(result) {
+  if (data.css) result.css = CHTMLSTYLES;
   if (!data.html && !data.htmlNode) return;
   var jax = MathJax.Hub.getAllJax()[0]; if (!jax) return;
   var script = jax.SourceElement(), html = script.previousSibling;
@@ -591,7 +592,6 @@ function GetHTML(result) {
   }
   if (data.html) result.html = html.outerHTML;
   if (data.htmlNode) result.htmlNode = html;
-  if (data.css) result.css = CHTMLSTYLES;
 }
 
 //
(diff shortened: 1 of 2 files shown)


dpvc:
I think the test should also check if the svg element is produced (one of the two changes relates to that).  I would also say that the tests should also check if *only* the desired properties are produced.  For example `speakText` appears in the `result` whether you have asked for it or not.

dpvc:
Other than that, the PR is good.

pkra:
Thanks! I'll update the test.

pkra:
> I would also say that the tests should also check if only the desired properties are produced. 

That will be covered by #285, correct?

dpvc:
> That will be covered by #285, correct?

It would be if that were implemented, yes.  But no one commented on it, so I was not sure if anyone thought that was a good idea.

pkra:
> But no one commented on it

Copy that. Just added +1.

Also, I've updated the test. PTAL?

dpvc:
OK, looks good.

PR #300, opened 2017-02-28
[README] add documentation section

### LINKED ISSUE #298: document config options
Last year, I created https://github.com/mathjax/MathJax-node/wiki/Configuration-options.

I'd like to propose to move into the readme (and update it in the process).

pkra:
Includes linting.
Resolves #298

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,9 +98,9 @@ The `start` method start (and restarts) mathjax-node. This allows reconfiguratio
 
 ### `typset(options, callback)`
 
-The `typeset` method is the main method of mathjax-node. It expects an configuration object `input` and a `callback`.
+The `typeset` method is the main method of mathjax-node. It expects a configuration object `input` and a `callback`.
 
-Once called, `typeset` can be called repeatedly and will optionally store information across calls (see `state` below); this information can be reset using the `start()` method (see above).
+Once called, `typeset` can be called repeatedly and will optionally store information across calls (see `state` below).
 
 The following are the default input options.
 
@@ -162,4 +162,4 @@ The `result` object will contain (at most) the following structure:
 }
 ```
 
-The `options` contains the configuration object passed to `typeset`; this can be useful for passing other data along.
+The `options` contains the configuration object passed to `typeset`; this can be useful for passing other data along or for identifying which `typeset()` call is associated with this `callback` call (in case you use the same `callback` function for more than one `typeset()`).
--- a/README.md
+++ b/README.md
@@ -1,41 +1,36 @@
 # mathjax-node [![Build Status](https://travis-ci.org/mathjax/MathJax-node.svg?branch=develop)](https://travis-ci.org/mathjax/MathJax-node)
 
-This repository contains a library that provides an API to call [MathJax](https://github.com/mathjax/mathjax) from
-NodeJS programs.  The API converts individual math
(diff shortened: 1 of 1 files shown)


pkra:
PTAL.

README.md
```diff
@@ -1,45 +1,40 @@
...
 
 mathjax-node v1.0 dropped the following features that were present in earlier pre-releases.
 
-* `lib/mj-page.js` (API for processing HTML-fragments) and related CLI tools
-* speech-rule-engine integration
-* PNG generation
+- `lib/mj-page.js` (API for processing HTML-fragments) and related CLI tools
+- speech-rule-engine integration
+- PNG generation
+- CLI tools in `bin/`
```
dpvc: I would add a comment here about the name change from `mj-single.js` to `main.js`, since people will have been loading that explicitly.

README.md
```diff
@@ -72,8 +67,95 @@ mjAPI.typeset({
...
+  state: {},                    // an object to store information from multiple calls (e.g., <defs> if useGlobalCache, counter for equation numbering if equationNumbers ar )
+  timeout: 10 * 1000,             // 10 second timeout before restarting MathJax
+}
+```
+
+### `callback(result, input)`
+
+mathjax-node returns two objects to the `callback`, a return object and the original configuration object.
+
+The result (at most) the following structure:
```
dpvc: Did the original `data` element ever get added to the returned result?  If so, it should be documented below.

pkra: Yes, it did (it's sort of mentioned in the heading but I forgot to write about it).

pkra:
I think I've addressed both (and tried my hand at having a more consistent language).

README.md
```diff
@@ -96,11 +98,11 @@ The `start` method start (and restarts) mathjax-node. This allows reconfiguratio
 
 ### `typset(options, callback)`
 
-The `typeset` method is the main method of mathjax-node. It expects a configuration object and a callback.
+The `typeset` method is the main method of mathjax-node. It expects an configuration object `input` and a `callback`.
```
dpvc: The change of 'a' to 'an' is not right, here.

README.md
```diff
@@ -96,11 +98,11 @@ The `start` method start (and restarts) mathjax-node. This allows reconfiguratio
...
 
-Once started, `typeset` can be called repeatedly and will potentially store information across calls (cf. `state` below).
+Once called, `typeset` can be called repeatedly and will optionally store information across calls (see `state` below); this information can be reset using the `start()` method (see above).
```
dpvc: Note that a `start()` call won't reset the `state`.  You would need to set `state` to an empty `{}` to do that.  You can do that without calling `start()`.  (Calling `start()` is good for allowing a new configuration to be used, however.)

README.md
```diff
@@ -159,3 +161,5 @@ The result (at most) the following structure:
          }
 }
 ```
+
+The `options` contains the configuration object passed to `typeset`; this can be useful for passing other data along.
```
dpvc: or for identifying which `typeset()` call is associated with this `callback` call (in case you use the same `callback` function for more than one `typeset()`

pkra:
PTAL @dpvc

dpvc:
OK, looks good.

PR #304, opened 2017-03-21
[main] improve passing of input data to output

### LINKED ISSUE #303: revisit passing data along to callback
While #239 was resolved, the original goal (to pass potentially unrelated data along) was not achieved.

The relevant code is https://github.com/mathjax/MathJax-node/blob/master/lib/main.js#L832-L835. The main problem seems to be that the input object is not actually passed to the callback but rather an intermediary object is built, based on the defaults (with values from the input data).

pkra:
Resolves #303

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -85,7 +85,7 @@ var timer;                           // used to reset MathJax if it runs too lon
 var document, window, content, html; // the DOM elements
 
 var queue = [];       // queue of typesetting requests of the form [data,callback]
-var data, callback;   // the current queue item
+var data, callback, originalData;   // the current queue item
 var errors = [];      // errors collected durring the typesetting
 var ID = 0;           // id for this SVG element
 
@@ -648,7 +648,7 @@ function GetSVG(result) {
 //  Start typesetting the queued expressions
 //
 function StartQueue() {
-  data = callback = null;       //  clear existing equation, if any
+  data = callback = originalData = null;       //  clear existing equation, if any
   errors = [];                  //  clear any errors
   if (!queue.length) return;    //  return if nothing to do
 
@@ -660,7 +660,7 @@ function StartQueue() {
   // and set the content with the proper script type
   //
   var item = queue.shift();
-  data = item[0]; callback = item[1];
+  data = item[0]; callback = item[1]; originalData = item[2];
   content.innerHTML = "";
   MathJax.HTML.addElement(content,"script",{type: "math/"+TYPES[data.format]},[data.math]);
   html.setAttribute("xmlns:"+data.xmlns,"http://www.w3.org/1998/Math/MathML");
@@ -775,7 +775,7 @@ function ReturnResult(result) {
     state.ID = ID;
   }
   serverState = STATE.READY;
-  callback(result, data);
+  callback(result, originalData);
   if (serverState === STATE.READY) StartQueue();
 }
 
@@ -836,7 +836,7 @@ exports.typeset = function (data,callback) {
   }}
   if (data.state) {options.state = data.state}
   if (!TYPES[options.format]) {ReportError("Unknown format: "+options.format,callback); return}
-  queue.push([options,callback]);
+  queue.push([options,callback,Object.assign({},data)]);
   if (serverState == STATE.STOPPED) {RestartMathJax()}
   if (serverState == STATE.READY) StartQueue();
 }
--- a/test/pass_data.js
+++ b/test/pass_data.js
@@ -1,0 +1,16 @@
+var tape = require('tape');
+var mjAPI = require("../lib/main.js");
+
+tape('Passing data along', function(t) {
+    t.plan(1);
+    mjAPI.start();
+    var tex = 'x';
+    mjAPI.typeset({
+        math: tex,
+        format: "TeX",
+        css: true,
+        something: 'expected',
+    }, function(data, input) {
+        t.equal(input.something, 'expected', 'Data was passed along to output');
+    });
+});


pkra:
Not sure if this is too simplistic.

dpvc:
I don't think this is the right approach, as it means that you are modifying the original data object (to add in the defaults), which is not a good idea, and since that object is part of the calling code, it could be modified outside of mathJax-node before mathjax-node has processed it.  That is one reason the mathjax-node currently copies the data rather than keep it directly.

Instead, I'd recommend adding a copy of the data object to the information stored in the queue (at [line 837](https://github.com/mathjax/MathJax-node/blob/a45cd48cca778c021ca13f4acbae78665c6ea1ac/lib/main.js#L837)), as

```
  queue.push([options,callback,Object.assign({},data)]);
```

and then at [line 662](https://github.com/mathjax/MathJax-node/blob/a45cd48cca778c021ca13f4acbae78665c6ea1ac/lib/main.js#L662) save the original as `originalData`:

```
 data = item[0]; callback = item[1]; originalData = item[2];
```

and return that at [line 776](https://github.com/mathjax/MathJax-node/blob/a45cd48cca778c021ca13f4acbae78665c6ea1ac/lib/main.js#L776)

```
  callback(result, originalData);
```

You will want to clear `originalData` at [line 650](https://github.com/mathjax/MathJax-node/blob/a45cd48cca778c021ca13f4acbae78665c6ea1ac/lib/main.js#L650), and define it originally at [line 88](https://github.com/mathjax/MathJax-node/blob/a45cd48cca778c021ca13f4acbae78665c6ea1ac/lib/main.js#L88).

pkra:
> I don't think this is the right approach, as it means that you are modifying the original data object (to add in the defaults), which is not a good idea, and since that object is part of the calling code, it could be modified outside of mathJax-node before mathjax-node has processed it. 

Right. But `Object.assign({},data)` give us a shallow copy as well, no? So we'd still have that risk, just somewhat reduced.

(Also, a separate PR may have been shorter :wink: )

dpvc:
> But `Object.assign({},data)` give us a shallow copy as well, no?

Yes, but the only nested structure that is part of the `data` object is the `state` object, and the is *supposed* to update as we go, so I'm not sure a copy of that is necessary.  In any case, the copy is not being used by mathjax-node; it is only kept in order to pass it back in the `callback`, so the risk to mathjax-node is entirely gone (it was the `options = data` that was the source of the risk).

If you want a deep copy, then use `Insert({},data)` instead.  I'm fine either way.

 > (Also, a separate PR may have been shorter 😉 )

but would not have constituted a code review, and would not have said why the changes were suggested, and would have required me to spend time testing it to make sure it works.  So it would have been shorter for *you*, but longer for *me*.  🙂

pkra:
@dpvc updated as per your suggestion. PTAL.

lib/main.js
```diff
@@ -659,7 +659,7 @@ function StartQueue() {
   //  and set the content with the proper delimiters
   //
   var item = queue.shift();
-  data = item[0]; callback = item[1];
+   data = item[0]; callback = item[1]; originalData = item[2];
```
dpvc: Can you remove the leading space that was added?

lib/main.js
```diff
@@ -835,7 +835,7 @@ exports.typeset = function (data,callback) {
   }}
   if (data.state) {options.state = data.state}
   if (!TYPES[options.format]) {ReportError("Unknown format: "+options.format,callback); return}
-  queue.push([options,callback]);
+    queue.push([options,callback,Object.assign({},data)]);
```
dpvc: Also here?

dpvc:
Other than the unwanted spaces, I think it is OK.

dpvc:
LGTM

lib/main.js
```diff
@@ -659,7 +659,7 @@ function StartQueue() {
   //  and set the content with the proper delimiters
   //
   var item = queue.shift();
-  data = item[0]; callback = item[1];
+  data = item[0]; callback = item[1]; originalData = item[2];
```
zorkow: Don't double park instructions.

lib/main.js
```diff
@@ -835,7 +835,7 @@ exports.typeset = function (data,callback) {
   }}
   if (data.state) {options.state = data.state}
   if (!TYPES[options.format]) {ReportError("Unknown format: "+options.format,callback); return}
-  queue.push([options,callback]);
+  queue.push([options,callback,Object.assign({},data)]);
```
zorkow: Object.assign is ES6. Is that intended?

pkra:
@zorkow Where's this coming from? This PR has already been reviewed by @dpvc so I wasn't planning any more change (just delaying for another bug fix release before a feature release with this).

> I think eventually they should be combined into a dedicated test runner, to avoid duplication and maintenance costs.

Please file an issue. (Though, as you know, I don't see the need at this point.)

> Don't double park instructions.

Let's please do a style clean-up some other time. A separate issue might be good.

> Object.assign is ES6. Is that intended?

Since Node v4 supports it (cf. the passing test), I don't see an issue.

pkra:
Ok, sorry for the mess of half-borked branches while working out that my local develop branch had been screwed up. 

This is now a clean new branch off develop with the same changes as an "earlier" commit.

I've now fixed two things I missed from @dpvc's earlier suggestion -- declaring and clearing `originalData`.

pkra:
@dpvc if you can take another look just for sanity's sake, I'd appreciate it. These are the same changes (plus two things I overlooked from your earlier recommendations).

dpvc:
LGTM

PR #305, opened 2017-03-21
Add test files from mathoid version of mathjax node

physikerwelt:
These tests have been created during the work on
the fork of the mathjax node.


pkra:
Thanks, Moritz, and sorry for the slow response. 

These tests are good but really about core MathJax, not mathjax-node. We plan to integrate them into the v3 test suite once the new TeX parser is ready.

PR #312, opened 2017-04-03
[readme] add documentation

### LINKED ISSUE #310: Undocumented properties of the `result` object
The documentation doesn't mention `result.errors` which seems to be pretty important.

There is also `result.width` and `result.height` that I'm finding very useful!

pkra:
document result.errors, result.height, result.width
Fixes #310


dpvc:
Looks good.

PR #313, opened 2017-04-03
fix #309

### LINKED ISSUE #309: sanitize user configuration [was: Cannot read property 'config']
Hello!
I am trying to use this module, with this settings:

```
mathR.config({
        displayMessages: true,
        displayErrors: true,
        undefinedCharError: true,
        MathJax: {
                jax: ["input/MathML","output/SVG"],
                extensions: ["mml2jax.js"]
        }
});
mathR.start();
```

and run with:

```
mathR.typeset({
      math: post.math,
      format: 'MathML',
                svg: true
},function (data) {if (!data.errors) console.log(data.svg);});

```

but it returns error:

```
/math-render/node_modules/mathjax-node/lib/main.js:682
  TEX.config.equationNumbers.autoNumber = data.equationNumbers;
     ^

TypeError: Cannot read property 'config' of undefined
```

What i am doing wrong?

pkra:
Fix #309

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -473,6 +473,8 @@ function ConfigureMathJax() {
   }
   if (MathJaxConfig) {
     adjustArrays(MathJaxConfig);
+    // ignore jax array from user config
+    delete MathJaxConfig.jax;
     Insert(window.MathJax,MathJaxConfig);
   }
 }
--- a/test/userconfig-jax.js
+++ b/test/userconfig-jax.js
@@ -1,0 +1,20 @@
+var tape = require('tape');
+var mjAPI = require('../lib/main.js');
+var jsdom = require('jsdom').jsdom;
+
+tape('User configuration with jax array', function (t) {
+        t.plan(1);
+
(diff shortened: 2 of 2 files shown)


dpvc:
See [my comment on #309](https://github.com/mathjax/MathJax-node/issues/309#issuecomment-291171592)

pkra:
The comments makes sense. However, I'd still be in favor of deleting the jax array because it seems to always be the right thing to do. 

The larger issue can still be fixed separately.

dpvc:
If you do the inserting that I suggest, that will overwrite the jax array (if specified in the user's configuration), so there is no need to delete it separately.  I don't see the need to fix one specific issue with a technique that will be obsolete if you make the global fix (and would want to be removed then).  That just makes doing the later fix more complicated, and one is likely to forget to remove the unneeded code at that point.  Unless you are not interested in fixing the issue in general, and just want to attack it piecemeal as people report the individual instances where it occurs, like this one.

PR #314, opened 2017-04-03
Fix #307

### LINKED ISSUE #307: [TeX input] autoload does not load the color extension [was: \definecolor does not work]
I'm not sure if that's a MathJax or a Mathjax-node bug. But it seems that \definecolor has stopped working

see
#306
for a test case

pkra:
Fix #307

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -422,7 +422,7 @@ function ConfigureMathJax() {
       //
       //  Reset the color extension after `autoload-all`
       //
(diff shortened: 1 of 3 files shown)


dpvc:
It is fine to add this check.  An additional change might be to not delete the `\textcolor`, `\definecolor`, and other macros, so that if they are used, the color extension *is* loaded.  But I suppose earlier uses of `\color` would then be inconsistent with the color extension.  Just a thought.

PR #317, opened 2017-04-11
sanitize user configuration

### LINKED ISSUE #309: sanitize user configuration [was: Cannot read property 'config']

pkra:
Fixes #309

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -473,7 +473,9 @@ function ConfigureMathJax() {
   }
   if (MathJaxConfig) {
     adjustArrays(MathJaxConfig);
-    Insert(window.MathJax,MathJaxConfig);
+    // merge the defaults into the user configuration (to sanitize)
+    var currentMathJax = Insert({},window.MathJax);
+    window.MathJax = Insert(Insert({},MathJaxConfig),currentMathJax);
   }
 }
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -145,9 +145,6 @@ function ConfigureMathJax() {
     jax: ["input/TeX", "input/MathML", "input/AsciiMath", "output/SVG", "output/CommonHTML"],
     extensions: ["toMathML.js"],
     TeX: {extensions: window.Array("AMSmath.js","AMSsymbols.js","autoload-all.js")},
-    tex2jax: {inlineMath: [['$','$'],['\\(','\\)']], preview:"none"},
-    mml2jax: {preview:"none"},
-    asciimath2jax: {preview:"none"},
     SVG: {useFontCache: true, useGlobalCache: false, EqnChunk: 1000000, EqnDelay: 0},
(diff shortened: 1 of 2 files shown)


pkra:
@dpvc PTAL.

lib/main.js
```diff
@@ -473,7 +473,9 @@ function ConfigureMathJax() {
   }
   if (MathJaxConfig) {
     adjustArrays(MathJaxConfig);
-    Insert(window.MathJax,MathJaxConfig);
+    // merge the defaults into the user configuration (to sanitize)
+    var currentMathJax = Insert({},window.MathJax);
+    window.MathJax = Insert(Insert({},MathJaxConfig),currentMathJax);
```
dpvc: These lines could be replaced by just

```
   window.MathJax = Insert(Insert({},MathJaxConfig),window.MathJax);
```

(I had originally intended to make the initial `window.MathJax = { ... }` that defines the mathjax-node configuration be a local variable until this point, but since this is inside an `if` statement, that would make things more complicated, so yours is better.)

I would also consider adding

```
   if (MathJaxConfig.extensions) {
     window.MathJax.extensions = window.MathJax.extensions.concat(MathJaxConfig.extensions);
   }
```

right after this, to append any user-provided extensions onto the mathjax-node ones.  I know that there is already `data.extensions` for that, but since the `extensions` array in the configuration is the usual way to do that, it might be good to handle the as well.

dpvc: When looking into this, I noticed that we no longer need to configure `tex2jax`, `mml2jax`, or `asciimath2jax`, since these are no longer used (we insert the explicit `<script>` tags now, rather than delimited math).  So lines 148 to 150 can be removed.

Along with that, the `Typeset` on line 701 can be changed to `Process`, since we don't need to do the preprocessor calls that are done by `Typeset()`.  Finally, the comment about setting the proper delimiters at line 661 could be changed to `//  and set the content with the proper script type`.

I think these could be part of this PR, but I know you like to do those things separately, so that could be done as a new PR if you want.

pkra: PTAL

zorkow:
LGTM. 
PS: I also like small PRs and my previous remarks on the tests applies here as well!

pkra:
@dpvc could you check that you're ok with this one?

Issue #318, opened 2017-04-11
Access MathJax.Hub in node?


beeplin:
mathjax-node only exposes methods such as `typeset`, `start` and `config`, but how to access `MathJax.Hub.xxx` in NodeJS?

beeplin:
Currently I found a workaround -- or, a hacking, by adding a new method `getMathJax` to `exports`:

```js
exports.getMathJax = function () {
  return MathJax;
}
```

And then in my node source code, call `mjAPI.getMathJax()` in the callback of the first calling of `mjAPI.typeset()`:

```js
mjAPI.typeset({}, () => { // empty typeset, only for initializing MathJax
  MathJax = mjAPI.getMathJax();
  MathJax.Hub.Register.StartupHook('xxx', () => {
    ...
  }
})
```
For Now it works fine, but I think it's better to design an better official API for doing this.

pkra:
> how to access MathJax.Hub.xxx in NodeJS?

I would avoid it if possible (since there might be negative side effects) but you can do this via the MathJax configuration, cf. http://docs.mathjax.org/en/latest/configuration.html#using-plain-javascript

```
   mjAPI.config = {
      MathJax: {
        AuthorInit: function () {
        MathJax.Hub.Register.StartupHook("xxx",function () {
          ...
        });
      }
      }
    }
```

beeplin:
that is just what I want~ thanks!

dpvc:
I don't think you can use `AuthorInit` in mathjax-node since mathjax-node sets `AuthorInit` itself in order to install the changes that it needs.  So setting this would wipe out mathjax-node's changes.

Can you be more specific about what your hook is going to do?  Perhaps there is another way to handle it.

beeplin:
Yeah I just figured out the `AuthorInit` isn't working. I am using `asciimath` and I need to define some new symbols in it, like this: 

```js
    MathJax.Hub.Register.StartupHook('AsciiMath Jax Ready', function () {
      AM = MathJax.InputJax.AsciiMath.AM
      AM.newsymbol({ input: '>=', tag: 'mo', output: '\u2A7E', tex: 'geqslant', ttype: AM.TOKEN.CONST })
      AM.newsymbol({ input: '<=', tag: 'mo', output: '\u2A7D', tex: 'leqslant', ttype: AM.TOKEN.CONST })
[...]

PR #319, opened 2017-04-13
[API change] exports.start() can accept a callback to get the MathJax object

beeplin:
In #318 we discussed how to access the `MathJax` object in mathjax-node. I suggest using the existing `start()` api to do this:
```js
//  E.g.
//     mjAPI.start(function (MathJax) {
//       MathJax.Hub.Register.StartupHook('XXX Ready', function () {
//         // something you need to run when XXX is ready
//       }
//     });
//
exports.start = function (callback) {
  RestartMathJax();
  exports.typeset({}, function () {
    callback && callback(window.MathJax);
  });
}
```

Currently the `start()` API is merely used in practice, because the first `typeset()` already does its job. Therefore I think it may be a reasonable design to make use of `start()` to access the `MathJax` object.


pkra:
Please make PRs based off develop (and to develop). Thanks!

PR #324, opened 2017-04-22
Update fontURL, support for reconfiguration

### LINKED ISSUE #323: update default fontURL
https://github.com/mathjax/MathJax-node/blob/master/lib/main.js#L41

pkra:
Fixes #323

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ The `config` method is used to set _global_ configuration options. Its default o
   displayErrors:   true, // determines whether error messages are shown on the console
   undefinedCharError: false, // determines whether "unknown characters" (i.e., no glyph in the configured fonts) are saved in the error array
   extensions: '', // a convenience option to add MathJax extensions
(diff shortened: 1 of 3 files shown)


pkra:
I ended up going with a simple `replace` in the `config()` method. Not sure if there are objections to that.

@dpvc is there neeed for `config()`to  be synchronized with the queue in general?

dpvc:
I don't think this change should be made.  As mentioned in #323, just as with MathJax in the browser, configuration should be done before any typesetting occurs.  There are many things that can't be reconfigured after typesetting begins; for example, you can't change the extensions after the initial typeset.  I would not complicate the code (and add maintenance costs) for one feature while leaving all the others unaddressed.  Is there really a use case for changing the font url between typesets?  If so, you can always do `mjAPI.start()` again after the reconfiguration.

So I recommend not applying this PR.

pkra:
See my comments on the issue.

pkra:
Closing in favor of #326

PR #343, opened 2017-07-13
v1.1.1 release

pkra:


test/output-html-linebreaks-manual.js
```diff
@@ -0,0 +1,17 @@
+var tape = require('tape');
+var mjAPI = require("../lib/main.js");
+var jsdom = require('jsdom').jsdom;
+
+tape('Output, HTML: linebreaks, manual', function(t) {
+  t.plan(1);
+  mjAPI.start();
+  var tex = 'A \\\\ B';
+  var expected = '<span class="mjx-chtml MJXc-display"><span class="mjx-math" style="width: 100%;" aria-label="A \\\\ B"><span class="mjx-mrow" style="width: 100%;" aria-hidden="true"><span class="mjx-stack" style="width: 100%; vertical-align: -1.2em;"><span class="mjx-block" style="text-align: center;"><span class="mjx-box"><span class="mjx-mi"><span class="mjx-char MJXc-TeX-math-I" style="padding-top: 0.519em; padding-bottom: 0.298em;">A</span></span></span></span><span class="mjx-block" style="text-align: center; padding-top: 0.467em;"><span class="mjx-box"><span class="mjx-mspace" style="width: 0px; height: 0px;"></span><span class="mjx-mi"><span class="mjx-char MJXc-TeX-math-I" style="padding-top: 0.446em; padding-bottom: 0.298em;">B</span></span></span></span></span></span></span></span>'
```
zorkow: Linebreaking.

lib/main.js
```diff
@@ -394,7 +394,7 @@ function ConfigureMathJax() {
               jax.CHTML.ex = ex = (data||defaults).ex;
               jax.CHTML.em = jax.CHTML.outerEm = em = ex / CHTML.TEX.x_height; // scale ex to x_height
               jax.CHTML.cwidth = width / em;
-              jax.CHTML.lineWidth = (linebreak ? width / em : CHTML.BIGDIMEN);
+              jax.CHTML.lineWidth = (linebreak ? width / em : 1000000);
```
zorkow: Why is that replaced. It does not differ from the value in chtml.
Either way, a magic number should be assigned to a named constant.

zorkow:
Changes rejected by author.

pkra:
These commits were all reviewed already. I don't think this is the time to change them.

lib/main.js
dpvc: @zorkow, in the CommonHTML file, this value *is* in a variable, but it is local to the file, and can't be accessed from outside, so we can't use it here.  The `CHTML.BIGDIMEN` value doesn't exist (and never did), and that is the source of the problem.  Since it is only used once in this location, I didn't see the need to name it in `main.js` when I suggested the change to Peter earlier.

PR #347, opened 2017-08-01
[main] typeset: add promise-based API

### LINKED ISSUE #297: Add a promise-based API
See also https://phabricator.wikimedia.org/T155201.

pkra:
Resolves #297

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -195,6 +195,10 @@ function ConfigureMathJax() {
         AddError("SVG - Unknown character: U+"+message[1].toString(16).toUpperCase()+
                     " in "+(message[2].fonts||["unknown"]).join(","),!undefinedChar);
       });
+      MathJax.Hub.Register.MessageHook("CommonHTML Jax - unknown char",function (message) {
+        AddError("CHTML - Unknown character: U+"+message[1].toString(16).toUpperCase()+
+                    " in "+(message[2].fonts||["unknown"]).join(","),!undefinedChar);
+      });
       MathJax.Hub.Register.MessageHook("MathML Jax - unknown node type",function (message) {
         AddError("MathML - Unknown node type: "+message[1]);
       });
--- a/lib/main.js
+++ b/lib/main.js
@@ -825,7 +825,9 @@ function RestartMathJax() {
 //     %%% cache results?
 //     %%% check types and values of parameters
 //
-exports.typeset = function (data,callback) {
+
+// callback API for compatibility with MathJax
+var cbTypeset = function (data, callback) {
   if (!callback || typeof(callback) !== "function") {
     if (displayErrors) console.error("Missing callback");
     return;
@@ -840,6 +842,17 @@ exports.typeset = function (data,callback) {
   if (serverState == STATE.STOPPED) {RestartMathJax()}
   if (serverState == STATE.READY) StartQueue();
 }
+
+// main API, callback and promise compatible
+exports.typeset = function (data, callback) {
+    if (callback) cbTypeset(data, callback);
+    else return new Promise(function (resolve, reject) {
+        cbTypeset(data, function (output, input) {
+            if (output.errors) reject(output.errors);
+            else resolve(output, input);
+        });
+    });
+};
 
 //
 //  Manually start MathJax (this is done automatically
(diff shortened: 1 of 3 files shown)


lib/main.js
```diff
@@ -841,6 +843,17 @@ exports.typeset = function (data,callback) {
   if (serverState == STATE.READY) StartQueue();
 }
 
+// main API, callback and promise compatible
+exports.typeset = function (data, callback) {
+    if (callback) cbTypeset(data, callback);
+    else return new Promise(function (resolve, reject) {
+        cbTypeset(data, function (output, input) {
+            if (output.errors) reject(output.errors);
```
dpvc: Do we need to be worried about the fact that `output.errors` an include warnings (like missing characters)?  There is a configuration option for turning this off, so it may not be necessary to worry about that here.

pkra: I would have said no since `undefinedCharError` is false by default (and if you switch it on, you likely want to reject the promise).

However, I have noticed that switching it to `true` has no effect for the promise-based API, e.g.,

```javascript
var mjAPI = require("./lib/main.js");
mjAPI.config({undefinedCharError: true});

mjAPI.typeset({
    math: "呵呵",
    format: "TeX",
    htmlNode: true,
    html: true,
    svg: true,
    equationNumbers: "AMS"
}, function (data) {
    console.log(data.errors); // has errors
});


mjAPI.typeset({math:'呵呵 ', html:true})
    .then((output, input) => console.log(output))
    .catch(errors => console.log("Errors:",errors)); //does not have errors
```

I don't understand what's going on here. (Hard errors work fine by the way, e.g., in the test file.)

dpvc: The problem turns out to be that the CHTML undefined character signals are not being trapped — only the SVG ones.  So your first `typeset()` call shows them because it has `svg: true`, but the second doesn't, because it is only for html.

You can fix this by adding

```
      MathJax.Hub.Register.MessageHook("CommonHTML Jax - unknown char",function (message) {
        AddError("CHTML - Unknown character: U+"+message[1].toString(16).toUpperCase()+
                    " in "+(message[2].fonts||["unknown"]).join(","),!undefinedChar);
      });
```

just after the corresponding SVG version at around line 198.

dpvc: > I would have said no since `undefinedCharError` is false by default

I thought it was true, but I see that you are right, so I agree, no need to do anything about it.  But at least we did find out that the CHTML undefined character signals aren't currently being trapped.  :-)

pkra:
@dpvc PTAL

test/base-warnings.js
```diff
@@ -0,0 +1,11 @@
+var tape = require('tape');
+var mjAPI = require("../lib/main.js");
+mjAPI.config({undefinedCharError: true});
+
+tape('basic test: check warnings', function (t) {
+    t.plan(2);
+    mjAPI.typeset({math:'呵', html:true})
```
dpvc: I'd prefer to see the math character as a unicode reference (`'\u5475`) rather than rely on the proper encoding of the javascript file.

test/base-warnings.js
```diff
@@ -0,0 +1,11 @@
...
+        .catch(errors => t.ok(errors, 'CommonHTML output reports error'));
+    mjAPI.typeset({math:'呵', svg:true})
```
dpvc: Same here.

dpvc:
Other than the raw unicode issue, this looks good, and is ready to merge.

test/base-warnings.js
pkra: One might argue that this won't be a problem on engines that understand promises but I'm happy to make the change.

pkra: edited the above.

pkra:
@dpvc PTAL.

test/base-warnings.js
dpvc: Because you can't specify the encoding used in a javascript source file, it is bad practice to use raw unicode (I assume this is UTF-8, but there are other unicode encodings).  For example, once of the editors I use will get this wrong, and were I to save it, the characters would change.

Were this to be loaded into a web page (and I know that it won't be), the javascript would be loaded (even in modern browsers, last time I checked) using the encoding of the web page itself, so if this were loaded into a non-UTF-8 page, it might not produce the correct character.  This is why we use explicit `\uXXXX` for the localization menu, for example.  It isn't about modern versus old browsers, it is about making assumptions about the encoding used.

PR #356, opened 2017-08-29
Fix fontURL issues

pkra:


lib/main.js
```diff
@@ -339,7 +339,9 @@ function ConfigureMathJax() {
         delete STYLES[".mjx-chartest .mjx-box"];
         delete STYLES[".mjx-test"];
         delete STYLES[".mjx-ex-boxtest"];
-
+        //  fontURL to current MathJax version
+        if (!fontURL)
+        fontURL = 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/' + MathJax.version + '/fonts/HTML-CSS';
```
dpvc: Probably should be indented, if not braced (Volker would vote for braces).

test/base-config-fonturl.js
```diff
@@ -10,12 +11,12 @@ tape('basic configuration: check fontURL', function (t) {
         format: "TeX",
         css: true
     }, function (result, data) {
-        t.ok(result.css.indexOf('https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/fonts/HTML-CSS') > -1, 'Default fontURL');
+        t.ok(result.css.indexOf('https://cdnjs.cloudflare.com/ajax/libs/mathjax/' + mjVersion + '/fonts/HTML-CSS') > -1, 'Default fontURL');
```
dpvc: The other option, I suppose, is to do a regex search with `mjVersion` replaced by something like `\d+\.\d+\.\d+`.  But I'm OK with the lookup as is.

lib/main.js
pkra: Oh, that should have been on the same line (otherwise we overwrite a custom fontURL, no?)

test/base-config-fonturl.js
pkra: Ah, nice idea. I somewhat favor this one but I'm ok either way.

lib/main.js
dpvc: No, it is still part of the `if` statement, even without braces.  The newline doesn't change the meaning.  Since it is a long expression, having it on a different line is fine.  Using braces would not be a bad idea (and will make Volker happy).

test/base-config-fonturl.js
dpvc: No problem.  Keep it as is, then.

lib/main.js
pkra: > will make @zorkow happy

Now how could I refuse that :smile:

pkra: oh, I apparently did not know that.

pkra:
@dvpc PTAL

dpvc:
LGTM

PR #357, opened 2017-08-29
[readme] document Promise API

### LINKED ISSUE #354: document new promise-based API
My bad - I forgot to do this as part of https://github.com/mathjax/MathJax-node/issues/297.

pkra:
Fixes #354

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,9 +96,11 @@ The `start` method start (and restarts) mathjax-node. This allows reconfiguratio
 
 **Note.** This is done automatically when `typeset` is first called (see below).
 
-### `typeset(options, callback)`
+### `typeset(options[, callback])`
 
-The `typeset` method is the main method of mathjax-node. It expects a configuration object `input` and a `callback`.
+The `typeset` method is the main method of mathjax-node. It expects a configuration object `input` and optionally a callback.
+
+If no `callback` is passed, it will return a Promise.
 
 Once called, `typeset` can be called repeatedly and will optionally store information across calls (see `state` below).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,9 +96,11 @@ The `start` method start (and restarts) mathjax-node. This allows reconfiguratio
 
 **Note.** This is done automatically when `typeset` is first called (see below).
 
-### `typeset(options, callback)`
+### `typeset(options[, callback])`
 
-The `typeset` method is the main method of mathjax-node. It expects a configuration object `input` and a `callback`.
+The `typeset` method is the main method of mathjax-node. It expects a configuration object `input` and optionally a callback.
```
dpvc: Shouldn't the configuration object be called `options` rather than `input`?

README.md
```diff
@@ -132,9 +134,9 @@ The following are the default input options.
 }
 ```
 
-### `callback(result, options)`
+### `Promise.resolve(result,options)` / `Promise.reject(errors) / `callback(result, options)`
```
dpvc: I think you need another backpack after `Promise.reject(errors)`.

README.md
```diff
@@ -165,4 +167,6 @@ The `result` object will contain (at most) the following structure:
 }
 ```
 
-The `options` contains the configuration object passed to `typeset`; this can be useful for passing other data along or for identifying which `typeset()` call is associated with this `callback` call (in case you use the same `callback` function for more than one `typeset()`).
+If the `errors` array is non-empty, the Promise will reject.
```
dpvc: How about:

    the Promise will reject, and be passed the `errors` array.

README.md
pkra: I didn't change it :smile:. I always think of it as input (since it contains the actual input). But I'm ok with changing it.

README.md
pkra: Thanks for catching that.

README.md
pkra: Sounds good.

README.md
dpvc: I know you didn't change it, but I did notice that it is not the same as the name above.  You could change `options` to `input` in the `typeset()` call if you prefer (and then update `options` elsewhere as well).

pkra:
@dpvc PTAL

dpvc:
LGTM

Issue #299, opened 2017-02-24
options for tweaking font fallback behavior


pkra:
From https://github.com/mathjax/MathJax-dev/issues/22#issuecomment-282081720

> MathJax-node can't measure the width of elements, like MathJax can in the browser. [...] So MathJax-node has to guess.  It does this by using the "monospace" font, and making an assumption about the width of that font.  For standard ASCII characters, this assumption is reasonable, but for non-western characters, this doesn't work so well. 

> You can adjust the width that MathJax-node uses by changing the `8.5` in [line 318](https://github.com/mathjax/MathJax-node/blob/master/lib/mj-single.js#L318) of `lib/mj-single.js` to be something larger; a value of 13 worked for me.  If you change the `monospace` to `serif` two lines later, then you might need to go to 15 or 16. 

I'm wondering: Should we expose this in the configuration options?

dpvc:
Fine by me, if you want to take time to do it and maintain it.

pkra:
This was fixed by #358.

PR #358, opened 2017-09-03
improve full-width characters rendering

ambar:


pkra:
Thanks for the PR.

Please note that PRs must be based on the `develop` branch (not master).

You also might want to add some comments on what you are trying to do. (I'm guessing this is for #299?)

ambar:
@pkra rebased!

ambar:
I tried to use Puppeteer(https://github.com/ambar/puptex) before I found this workaround, do you think it's worth?

pkra:
@ambar a fix for #299 would be great (and should have better overall performance than spinning up puppeteer/Chrome). But @dpvc will have to comment on the approach you're taking.

ambar:
@pkra Thanks!

dpvc:
@pkra, I'm OK with the change, if you want to merge it.

pkra:
Thanks. I just haven not got around to it -- the pending bug fix release needs to be wrapped up before we merge.

pkra:
Sorry for taking so long to come back to this.

We're ok with merging as is. 

(We thought about adding an option to customize widths of halfwidth forms but that can be done later.)

pkra:
I'm wondering if we should generalize this using https://github.com/sindresorhus/string-width. I'll make a new issue.

Issue #325, opened 2017-04-26
jsdom 10 compatibility


pkra:
jsdom 10 has changed its API.

It's also node 6+ only now.

pkra:
@dpvc any chance you can look into this some time?

dpvc:
I've submitted a PR for this.

dpvc:
Looks like I need to fix some tests first.  Sorry about that.  I'm stuck on one with mglyph, since jsdom10 doesn't have the same flexibility about blocking the loading of items as version 9 did (you could allow scripts, but not images, and that can no longer be done, apparently).  So I need to override mode of MathJax to get mglyph to work properly with jsdom 10.

pkra:
:heart: Thanks, Davide.

pkra:
jsdom 10+ is node6+ only, so this should cause a major version jump, I think.

pkra:
Oh, and travis needs to be restricted as node4 and 5 probably always fail.

dpvc:
> Oh, and travis needs to be restricted as node4 and 5 probably always fail.

I don't know how travis is set up or configured.  Can you point me to the right place to do that?

pkra:
Sure! It's https://github.com/mathjax/MathJax-node/blob/master/.travis.yml#L3-L4

dpvc:
Thanks.  I didn't see it because it was a dot file, and I didn't think to look for that (am used to .json files now)

dpvc:
OK, the tests all pass now.  I think it is ready to go.

dpvc:
I've updated the version number and added the node6 engine dependency.

dpvc:
==> Merged

PR #376, opened 2018-01-05
Update to jsdom v10 interface.  Resolves issue #325.

dpvc:
This PR uses the new jsdom10 API (rather than the older jsdom9 API).  As part of the update, it removes tests from the jsdom patch file that no longer are needed.

Resolves issue #325.


pkra:
I can't vouch for the slimmed down jsdom patch (but will happily take your word for it :smile:); otherwise LGTM.

Issue #378, opened 2018-01-15
SVG unknown char messages when SVG output is not set


pkra:
Originally reported at https://github.com/pkra/mathjax-node-page/issues/56.

The following will produce `"SVG - Unknown character: U+C in MathJax_Main,MathJax_Size1,MathJax_AMS"` etc.

```
var mjAPI = require("mathjax-node");
mjAPI.typeset({
  math: 'a = \frac{\text{some text with umlauts äöü}}{1.23}',
  format: "TeX", // "inline-TeX", "MathML"
  mml:true,
}, function (data) {
  if (!data.errors) {console.log(data.mml)}
});
```

On the other hand, there are no MML Unknown Char messages so I'm guessing the message hooks are not set up correctly. Maybe they should just be independent of the output anyway?

dpvc:
I've made a pull request to fix the issue.

dpvc:
==> Merged.

PR #379, opened 2018-01-15
Use new None output for mml-only output

dpvc:
Resolves issue #378


pkra:
LGTM

dpvc:
I'm thinking about adding a do-nothing output renderer and using that for this case instead.  Another use case would be for input syntax checking, where the output isn't really needed.

dpvc:
I've added a `None` output jax that does nothing and can be used in the case where just MathML output is needed (since the `toMathML()` function is used for that, not the output of the NativeMML output jax), or when just syntax checking of the input is needed.

Issue #134, opened 2015-12-29
loading third party extensions


pkra:
Just a note to think about how users can specify third party extensions (i.e., without copying extensions into the right folder of core MathJax).

pkra:
I'm thinking a simple configuration option that can be used to specify a folder (where developers store their third party extensions) should work ok. That folder can be resolved and set as `Contrib` path in core MathJax.

I'm not sure how jsdom behaves in terms of security (since mathjax-node works on `fs://`).

dpvc:
You may not want to override `[Contrib]` but rather make a new path for these.  But yes, a way to do that would be useful.

pkra:
> You may not want to override `[Contrib]` but rather make a new path for these

My reasoning was 

a) the hard-coded `[Contrib]` path for the CDN's contrib folder shouldn't work in mathjax-node -- I had actually assumed we disallow remote sources when opening the jsdom window (for security reasons) but I see now that's not the case; I'll create an issue.
b) re-using `[Contrib]` makes it easy to use extension from the third party extensions repo.

dpvc:
>  the hard-coded `[Contrib]` path for the CDN's contrib folder shouldn't work in mathjax-node

I'm not sure why not.

> re-using `[Contrib]` makes it easy to use extension from the third party extensions repo.

Using another path would not make it any harder.  You will have to give an explicit path to the extension anyway, so I'm not sure why `[OtherPath]/extension.js` is harder than `[Contrib]/extensino.js`.  The only extensions that actually use `[Contrib]` internally are the a11y extensions, and they use `[a11y]` for their paths, so configuring that would make them work even if they are loaded from `[OtherPath]`.  So I don't see what is gained, really.  Not overriding `[Contrib]` allows extensions to be loaded directory from the CDN; overriding it would force them to be installed locally (more work and chance of error).

pkra:
> Using another path would not make it any harder. Y

Maybe I still don't understand the internals then. 

All third party extensions use `MathJax.Ajax.loadComplete("[Contrib]/...");`. Wouldn't a mathjax-node user have to change that?

dpvc:
Ah yes, you are correct.  I had forgotten about that.  `[Contrib]` it is, then.

pkra:
Moving this out of the milestone.

pkra:
Since the CDN was retired, third party extensions should now have their own prefix so this should be easier.

pkra:
[...]

PR #388, opened 2018-02-08
387

pkra:


pkra:
I'm not sure why two commits are listed. I had accidentally opened this to master, then closed, then switched to develop and re-opened.

dpvc:
> I'm not sure why two commits are listed.

It's probably because I messed up the travis change and merged into master, and then had to do the same to develop.  The merge to develop technically isn't in master; even though it is the same change, the merge is a new commit.

I will make the 2.0.2 release once MathJax v2.7.3 is out, updating the package.json to call in the new version as well.

pkra:
> I will make the 2.0.2 release once MathJax v2.7.3 is out, updating the package.json to call in the new version as well.

Thanks. I have a fix for #134 in the works but that can wait until after 2.0.2 (since it would bump to 2.1.0, right?)

dpvc:
Yes, I think that is right.

test/issue219.js
```diff
@@ -20,7 +20,7 @@ tape('Basic Check: pass jsdom object to output', function(t) {
   }, function(data) {
     var window = new JSDOM().window;
     t.ok(data.htmlNode instanceof window.HTMLElement, 'htmlNode is an HTMLElement');
-    t.ok(data.svgNode instanceof window.HTMLElement, 'svgNode is an HTMLElement');
-    t.ok(data.mmlNode instanceof window.HTMLElement, 'mmlNode is an HTMLElement');
+    t.ok(data.svgNode instanceof window.SVGElement, 'svgNode is an HTMLElement');
+    t.ok(data.mmlNode instanceof window.Element, 'mmlNode is an HTMLElement');
```
dpvc: Shouldn't these read `svgNode is an SVGElement` and `mmlNode is an Element` instead of `HTMLElement` for both?

pkra: Yes. But it's merged now?

dpvc: Yes, I was adding this to remind myself to fix it in a new PR.

pkra: Thanks for clarifying.

dpvc: I've taken care of this now.

PR #390, opened 2018-02-09
[package] update MathJax to 2.7.3

### LINKED ISSUE #389: update to mathjax v2.7.3
Since mj-node hard-codes the version, this needs upgrading upon release.

pkra:
Resolves #389

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -584,9 +584,9 @@
       "integrity": "sha1-7dFMgk4sycHgsKG0K7UhBRakJDg="
     },
     "mathjax": {
-      "version": "2.7.2",
-      "resolved": "https://registry.npmjs.org/mathjax/-/mathjax-2.7.2.tgz",
-      "integrity": "sha512-Z9te7r3lsjZibugO1uKsdyqICKXVNr7M1Ll2GtjJu9cUKvOvwDqEp0YIjBD4H58NNuPg5DP5/AQ2Tu8K52Mqng=="
+      "version": "2.7.3",
+      "resolved": "https://registry.npmjs.org/mathjax/-/mathjax-2.7.3.tgz",
+      "integrity": "sha512-HEKJYRofOD5k9RmTj1amdA32JIJOqJ8SO+lq50mffZzdfFz+bNnsB0tJ8efMfYpiffvUHkqFRZFxXftW4BaBOw=="
     },
     "mime-db": {
       "version": "1.30.0",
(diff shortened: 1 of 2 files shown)


dpvc:
I'm not sure this is necessary, since the current version is `^2.7.2`, which should match any version `2.x,y` that is at least 2.7.2.  So 2.7.3 should match already.

This is supposed to represent a *minimim* version, right?  So I'm wondering if 2.7.2 is even too high, since it works with 2.7.0 (and probably 2.6.0).

Similarly, we changes jsdom to version 11, but it does work with version 10, so shouldn't the version be `10.0.0-11.*.*`?  If someone is using jsdom as part of a larger project, and is at version 10.6.2, and wanted to use mathjax-node as well, then we would be forcing them to update to jsdom 11 even if they weren't ready to do that yet.

While the mathjax version probably isn't going to cause trouble, something like jsdom certainly could.

pkra:
> I'm not sure this is necessary, since the current version is ^2.7.2, which should match any version 2.x,y that is at least 2.7.2. So 2.7.3 should match already.

Ah, I had overlooked the caret and thought it was hardcoded. You are right.

> This is supposed to represent a minimim version, right? So I'm wondering if 2.7.2 is even too high, since it works with 2.7.0 (and probably 2.6.0).

I would not drop below 2.7.2 because of the Safari regression that was fixed with an additional font file in 2.7.2.

> Similarly, we changes jsdom to version 11, but it does work with version 10, so shouldn't the version be 10.0.0-11.*.*?

I don't mind either way.

> If someone is using jsdom as part of a larger project, and is at version 10.6.2, and wanted to use mathjax-node as well, then we would be forcing them to update to jsdom 11 even if they weren't ready to do that yet.

I believe it would only lead to both versions being installed (with v11 being isolated within mathjax-node's module folder). But yes if someone wants to have mathjax-node and jsdom v10 interact, they might theoretically have trouble (though again the breaking changes in v11 seem rare).

pkra:
So I guess this can be closed?

dpvc:
> I would not drop below 2.7.2 because of the Safari regression that was fixed with an additional font file in 2.7.2.

OK.

> I believe it would only lead to both versions being installed 

Probably right.  Perhaps it would not be a problem.  Not sure if it needs changing or not, then.  I'll think about it some more.

dpvc:
> So I guess this can be closed?

Yes, I will close it.

Issue #392, opened 2018-02-27
useFontCache:false has no effect


pkra:
[Reported on SO](https://stackoverflow.com/questions/48997659/trying-to-use-usefontcache-false-in-mathjax-node).

```
var mjAPI = require("mathjax-node");

mjAPI.config({MathJax: {SVG: {font: "TeX"}}});
mjAPI.start();

mjAPI.typeset({
  math: "\sin^2+\cos^2=1",
  format: ("TeX"),
  svg:true,
  useFontCache: false,
  ex: 6, width: 100,
}, function (data) {
  if (!data.errors) {console.log(data.svg)}
});
```

results in 

```
<svg xmlns:xlink="http://www.w3.org/1999/xlink" width="15.722ex" height="2.843ex" style="vertical-align: -0.505ex;" viewBox="0 -1006.6 6769.3 1223.9" role="img" focusable="false" xmlns="http://www.w3.org/2000/svg" aria-labelledby="MathJax-SVG-1-Title">
<title id="MathJax-SVG-1-Title">sin^2+cos^2=1</title>
<defs aria-hidden="true">
<path stroke-width="1" id="E1-MJMATHI-73" d="M131 289Q131 321 147 354T203 415T300 442Q362 442 390 415T419 355Q419 323 402 308T364 292Q351 292 340 300T328 326Q328 342 337 354T354 372T367 378Q368 378 368 379Q368 382 361 388T336 399T297 405Q249 405 227 379T204 326Q204 301 223 291T278 274T330 259Q396 230 396 163Q396 135 385 107T352 51T289 7T195 -10Q118 -10 86 19T53 87Q53 126 74 143T118 160Q133 160 146 151T160 120Q160 94 142 76T111 58Q109 57 108 57T107 55Q108 52 115 47T146 34T201 27Q237 27 263 38T301 66T318 97T323 122Q323 150 302 164T254 181T195 196T148 231Q131 256 131 289Z"></path>
[...]

PR #393, opened 2018-02-27
Define the None output jax after the other jax are loaded

dpvc:
Define the None output jax after the other jax are loaded (so SVG will be loaded initially, so its configuration can be set by hand).  Resolves issue #392.


pkra:
LGTM.

pkra:
*ping*

PR #402, opened 2018-04-11
 [main] extensions options: support names containing "s"

### LINKED ISSUE #401: [main] extensions options breaks with 2+ extensions
This seems to be due to

https://github.com/mathjax/MathJax-node/blob/a3c71820b5c0cf0848a48d86df9828aed9c796d2/lib/main.js#L531

missing an escape -- I suspect it should be `var extensionList = extensions.split(/\s*,\s*/);`.

pkra:
Fixes #401

### DIFF
--- a/test/base-01-errors-startup.js
+++ b/test/base-01-errors-startup.js
@@ -1,0 +1,28 @@
+var tape = require('tape');
+var mjAPI = require('../lib/main.js');
+
+// NOTE this test must first (in tape) to test the startup phase (hence the filename contains "01")
+
+tape('Catch errors during startup phase', function(t) {
+  t.plan(1);
+  mjAPI.config({
+    extensions: 'blargh'
+  });
+  mjAPI.typeset(
+    {
+      math: 'x',
+      format: 'TeX',
+      mml: true
+    },
+    function(data) {
+      t.ok(
+        data.errors,
+        'Error (loading non-existent extension) is caught in output'
+      );
+     // reset configuration
+     mjAPI.config({
+        extensions: 'tex2jax'
+      });
+    }
+  );
+});
--- a/lib/main.js
+++ b/lib/main.js
@@ -532,7 +532,7 @@ function ConfigureMathJax() {
     //
     // Parse added extensions list and add to standard ones
     //
-    var extensionList = extensions.split(/s*,\s*/);
+    var extensionList = extensions.split(/\s*,\s*/);
     for (var i = 0; i < extensionList.length; i++) {
       var matches = extensionList[i].match(/^(.*?)(\.js)?$/);
       window.MathJax.extensions.push(matches[1] + '.js');
--- a/lib/main.js
+++ b/lib/main.js
@@ -91,6 +91,7 @@ var document, window, content, html; // the DOM elements
 var queue = [];       // queue of typesetting requests of the form [data,callback]
 var data, callback, originalData;   // the current queue item
 var errors = [];      // errors collected durring the typesetting
+var sErrors = [];     // errors collected durring MathJax startup
 var ID = 0;           // id for this SVG element
 
 //
@@ -519,7 +520,10 @@ function ConfigureMathJax() {
         if (MathJax.OutputJax.SVG.resetGlyphs) MathJax.OutputJax.SVG.resetGlyphs(true);
         MathJax.ElementJax.mml.ID = 0;
         serverState = STATE.READY;
-        MathJax.Hub.Queue(StartQueue);
+        MathJax.Hub.Queue(
+            function () {sErrors = errors},
+            StartQueue
+          );
       });
     }
   };
@@ -730,7 +734,7 @@ function GetSVG(result) {
 //
 function StartQueue() {
   data = callback = originalData = null;       //  clear existing equation, if any
-  errors = [];                  //  clear any errors
+  errors = sErrors; sErrors = [];              //  clear any errors
   if (!queue.length) return;    //  return if nothing to do
 
   serverState = STATE.BUSY;
--- a/test/base-config-extensions.js
+++ b/test/base-config-extensions.js
@@ -1,0 +1,22 @@
+var tape = require('tape');
+var mjAPI = require('../lib/main.js');
+
+tape('Options "extension": multiple extensions', function(t) {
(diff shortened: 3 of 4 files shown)


pkra:
Oh! I didn't push the actual commit. Oh well, might as well check the test for this part first.

pkra:
The failing test are about the fontURL option. I think those tests should be expected to fail on old node/npm versions since the test relies on `package-lock.json`.

test/base-01-errors-startup.js
```diff
@@ -0,0 +1,28 @@
+var tape = require('tape');
+var mjAPI = require('../lib/main.js');
+
+// NOTE this test must first (in tape) to test the startup phase (hence the filename contains "01")
```
dpvc: I think if you use `mjAPI.start()` after the config in your callback, you should be able to put this test anywhere, as the should restart MathJax (by obtaining a whole new jsdom window).

dpvc:
(Sorry, hit the wrong button).

test/base-01-errors-startup.js
pkra: copy that.

dpvc:
> I vaguely recall that the problem might resolve by updating the package-lock so as to have mathjax 2.7.4. I'll try that locally.

Maybe just have it look for any version in the string rather than look up the specific one.  Or maybe you could load the mathjax module and get the version directly.

Something like

```
window = {};
window.MathJax = MathJax = {};
document = {getElementById: function () {}, childNodes: [], createElement: function () {}};
try {require("mathjax")} catch (e) {};
console.log(MathJax.version);
```

should get you the version of MathJax currently installed.  It's a bit of a hack, but might work better than reading the package-lock file.

pkra:
So we have three options

* skip the test on old node versions (by checking `process.version)
* update package-lock to have mathjax 2.7.4
* switch to the workaround in https://github.com/mathjax/MathJax-node/pull/402#issuecomment-380524715

@dpvc and @zorkow, how would you want me to resolve this?

(Since I think the package-lock file should be updated to 2.7.4 for consistency, I'd favor this approach but it's not my call.)

dpvc:
It seems to me that relying on a file (`package-lock.json`) that is not maintained in node 6 and 7 means that the test is broken.  Even though the `package-lock.json` file has 2.7.3, if npm installs 2.7.4, it will update the version in the file automatically in versions of npm that understand the package-lock file (node 8 and above).  Relying on package-lock in the other version is essentially the same as hard-coding the version number into the test file (and that would at least be more honest).

If we want to maintain tests for node 6 and 7, then I don't think we can rely on the package-lock version, since it need not be the actual version installed.  If you want to run these tests in node 6 and 7, we need an alternative method of finding MthJax's version.  I offered one above, but have worked out another one, which you can see in this [comparison](https://github.com/mathjax/MathJax-node/compare/develop...version-detect?expand=1).  Here, we use an extension that defines a TeX macro that returns the MathJax version, and then extract the version from the resulting output.  The prevents having to polite the global namespace (as in my other approach).

This should work in all versions of node, and with whatever version of MathJax is loaded.  That would be my recommendation.  If that seems acceptable, I will make a separate pull request, and we can merge it prior to this one, and I think that will take care of the build fails without needing further modification to this PR.

test/base-01-errors-startup.js
```diff
@@ -0,0 +1,28 @@
...
+      mml: true
+    },
+    function(data) {
+      t.ok(
+        data.errors,
+        'Error (loading non-existent extension) is caught in output'
+      );
+     // reset configuration
+     mjAPI.config({
+        extensions: 'tex2jax'
```
dpvc: I don't think this is a reset, actually. The `text2jax` extension is not loaded by mathjax-node, so this is an alteration.  I think you want

```
extensions: ''
```

since the default extensions string is empty.

test/base-01-errors-startup.js
pkra: Would it be best to restart after resetting the config at the end?

test/base-01-errors-startup.js
```diff
@@ -0,0 +1,28 @@
...
+  mjAPI.typeset(
+    {
+      math: 'x',
+      format: 'TeX',
+      mml: true
+    },
+    function(data) {
+      t.ok(
+        data.errors,
+        'Error (loading non-existent extension) is caught in output'
```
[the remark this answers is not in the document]
pkra: That had no effect for me and I couldn't figure out how to avoid the virtualconsole's errors coming through.

test/base-01-errors-startup.js
pkra: Copy that.

pkra:
Re https://github.com/mathjax/MathJax-node/pull/402#issuecomment-381288082

Copy that. I've made a PR #404 into this branch to get things started.

pkra:
To recap, I've updated the test to fix the restart but I couldn't get the errors to hide from console.

I've also made a suggestion in the other PR to fix the test.

dpvc:
Here is yet *another* way to get the version (and to get access to the `MathJax` object, which I hadn't thought was possible, but is).

```
mjAPI.typeset({
  math: '',
  format: "TeX",
  htmlNode: true,
}, function (data) {
  console.log(data.htmlNode.ownerDocument.defaultView.MathJax.version);
});
```

Perhaps that is better than the external module, since that doesn't require a configuration change, or external file?

pkra:
> Perhaps that is better than the external module, since that doesn't require a configuration change, or external file?

If you prefer this, I'm happy to update the test.

dpvc:
I think merging the develop branch should now allow your PR to pass the tests.  I've fixed a timing issue that was causing problems when tests were run sequentially even when they worked individually.

dpvc:
The conflict listed in `lib/main.js` can be resolved by changing

```
MathJax.Hub.Queue(StartQueue);
```

to 

```
MathJax.Hub.Queue(
    function () {sErrors = errors},
    StartQueue
);
```

in the `develop` version and removing the  `401` version.

pkra:
@dpvc could you review the conflict resolution?

dpvc:
Looks good.  Sorry for the long run around to get what in the end is a pretty small set of changes...

Issue #163, opened 2016-02-10
integrate PreviewHTML


pkra:
[This has been separated from #111. ]

If there is sufficient interest, we could enable the PreviewHTML in MathJax-node.

The PreviewHTML an HTML/CSS output that's lower-quality (both in terms of supported constructions and quality) than the CommonHTML; it's basically the old version of the CommonHTML from MathJax v2.5.  

The advantages would be smaller HTML fragment size, the fact that it is optimized for TimesNewRoman or similar local fonts (i.e., no webfonts) and rendering speed.

dpvc:
It is not feature complete, however, so some things don't render.  E.g., menclose is not implemented, and I think horizontal stretchy characters may not be, and the spacing can be a bit wonky.

pkra:
Cleaning up my old issues.

PR #403, opened 2018-04-17
[WIP] PreviewHTML support

pkra:
A first proposal towards #163. 

Looking for feedback.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -824,10 +824,10 @@ function StartQueue() {
   //
   // Get the renderer to use
   //
-  var renderer = (
-      (data.html || data.htmlNode || data.css) ? "CommonHTML" :
-          (data.svg || data.svgNode) ? "SVG" : "None"
-  );
+  var renderer = 'None';
+  if (data.html || data.htmlNode || data.css) renderer = "CommonHTML";
+  else if (data.svg || data.svgNode) renderer = "SVG"
+  else if (data.phtml || data.phtmlNode || data.pcss) renderer = "PreviewHTML"
 
   //
   //  Set up a timeout timer to restart MathJax if it runs too long,
--- a/lib/main.js
+++ b/lib/main.js
@@ -58,7 +58,10 @@ var defaults = {
 
   html: false,                    // return HTML output?
   htmlNode: false,             // DOM element for HTML output?
+  phtml: false,                    // return pHTML output?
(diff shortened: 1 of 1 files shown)


pkra:
I've rebased and all tests pass now.

@dpvc could you take a quick look and tell me if this approach is generally ok with you? (I had a quick chat with @zorkow who suggested your review).

Then I could start to write a few basic tests.

dpvc:
It seems to be doing the right thing.  I haven't looked through the PreviewHTML code to see what else would need to be overriden, but the framework seems to be OK.

I'm not sure what the use case is, however.  The preview HTML results are not very good, and it is not feature complete.  In general, it is not output that I would like to see being used in production work as the final layout.

pkra:
@dpvc ping - if you have some time.

lib/main.js
```diff
@@ -780,10 +824,10 @@ function StartQueue() {
...
   // Get the renderer to use
   //
-  var renderer = (
-      (data.html || data.htmlNode || data.css) ? "CommonHTML" :
-          (data.svg || data.svgNode) ? "SVG" : "None"
-  );
+  var renderer = 'None';
+  if (data.html || data.htmlNode || data.css) renderer = "CommonHTML";
+  else if (data.svg || data.svgNode) renderer = "SVG"
+  else if (data.phtml || data.phtmlNode || data.pcss) renderer = "PreviewHTML"
```
dpvc: I think you need to swap lines 829 and 830, since you are checking for PHTML before SVG below (and the serenader routines assume the preference order is CHTML > PHTML > SVG).

lib/main.js
```diff
@@ -882,8 +928,20 @@ function SetRenderer(renderer) {
   return MathJax.Hub.setRenderer(renderer);
 }
 
+function RerenderPHTML(result) {
+  if ((data.html || data.htmlNode || data.css) && (data.phtml || data.phtmlNode || data.pcss)) {
```
dpvc: I'm wondering if you could just check the `renderer` variable here:

```
if (renderer !== 'PreviewHTML' && (data.phtml || data.phtmlNode || data.pcss)) {
  renderer = 'PreviewHTML';
...
}
```

and make a similar change to the `RerenderSVG()` function below.

pkra:
Thanks. As discussed F2F, this won't be accepted.

PR #405, opened 2018-04-18
Fix start() to handle being called while MathJax is starting up

dpvc:
If `start()` was called while MathJax was starting up from a previous `start()` or `typeset()` call, this could cause crashes (and did cause crashes for some tape tests). 

This PR fixes `start()` so that it delays the restart until the previous MathJax has started up.

It also removes unneeded `start()` calls from the test files (no need to create a new window and reload MathJax unnecessarily), and adds configuration resets for those tests that include configuration changes.  Finally, it normalizes the format of the tests.


dpvc:
This also fixes the version check so that it works in all node and mathjax versions.

PR #412, opened 2018-05-31
Rc 2.1.1

dpvc:
This is the release candidate for 2.1.1.  It fixes the extensions `s` issue, and includes a fix to `start()` method that avoids a timing issue if it is called twice in succession, and traps errors during the startup.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "mathjax-node",
-  "version": "2.1.0",
+  "version": "2.1.1",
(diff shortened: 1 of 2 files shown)


pkra:
Loooks good but won't `develop` end up missing the last commit?

dpvc:
>  won't `develop` end up missing the last commit?

Sorry, this was supposed to be to `develop` (then one to `master` following that).  I could have sworn I had changed that.  Oh, well, done now.

PR #425, opened 2018-09-06
bump package-lock to MathJax 2.7.5

pkra:


pkra:
This will get rid of the vulnerability warning for MathJax.

pkra:
This commit should probably be cherry picked into master as well to get rid of the warning when viewing the repo on github.

dpvc:
> This commit should probably be cherry picked into master as well to get rid of the warning when viewing the repo on github.

Technically, it should have been done as a hot-fix on the master branch, and then the same commit could be pulled into develop.  But I'll do the cherry pick to handle it as is.

Issue #441, opened 2019-06-12
Uncaught [TypeError: Cannot read property 'setAttribute' of undefined]


connerbw:
I'm working on an express.js app that wraps this library and returns a PNG image. Sometimes the equations will be less than stellar, possibly garbage.  This example crashes my node app, leaves it in an usable state, I have to restart the server:

## Equation: 

`\begin{equation*}  A = \left[ \begin{array}{cccc}  a_{11} & a_{12} & a_{13} & a_{14} \\  a_{21} & a_{22} & a_{23} & a_{24} \\  a_{31} & a_{32} & a_{33} & a_{34}  \end{array} \right]  \end{equation*}`

## Error:

```
Math Processing Error: Cannot read property 'toFixed' of undefined                                                                                                      
Error: TypeError: Cannot read property 'setAttribute' of undefined                                                                                                      
    at GetSVG (C:\Users\dac\Code\github\pressbooks\pb-mathjax\node_modules\mathjax-node\lib\main.js:698:7)                                                              
    at Function.execute (file:///C:/Users/dac/Code/github/pressbooks/pb-mathjax/node_modules/mathjax/unpacked/MathJax.js:243:26)                                        
    at cb (file:///C:/Users/dac/Code/github/pressbooks/pb-mathjax/node_modules/mathjax/unpacked/MathJax.js:225:59)                                                      
    at Object.Process (file:///C:/Users/dac/Code/github/pressbooks/pb-mathjax/node_modules/mathjax/unpacked/MathJax.js:495:38)                                          
    at Object.call (file:///C:/Users/dac/Code/github/pressbooks/pb-mathjax/node_modules/mathjax/unpacked/MathJax.js:508:37)                                             
    at Function.WAITEXECUTE (file:///C:/Users/dac/Code/github/pressbooks/pb-mathjax/node_modules/mathjax/unpacked/MathJax.js:348:50)                                    
    at cb (file:///C:/Users/dac/Code/github/pressbooks/pb-mathjax/node_modules/mathjax/unpacked/MathJax.js:225:59)                                                      
    at Object.Process (file:///C:/Users/dac/Code/github/pressbooks/pb-mathjax/node_modules/mathjax/unpacked/MathJax.js:495:38)                                          
    at Object.call (file:///C:/Users/dac/Code/github/pressbooks/pb-mathjax/node_modules/mathjax/unpacked/MathJax.js:508:37)                                             
[...]

PR #442, opened 2019-06-13
Make sure SVG and HTML elements exist before using them.  #441

dpvc:
If the input jax fails to process the given input, the resulting SVG or HTML element for the output will not exist.  So check that svg and html elements exist before trying to use them.

This resolves issue #441.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -620,6 +620,7 @@ function AddError(message,nopush) {
 function GetMML(result) {
   if (!data.mml && !data.mmlNode) return;
   var jax = MathJax.Hub.getAllJax()[0];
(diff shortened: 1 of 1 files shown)


pkra:
This looks ok to me.

But perhaps there should be some error or warning when an output is requested but cannot be fulfilled?

dpvc:
When the SVG or HTML element is not generated, that means something went wrong in the input our output jax, and that should cause a `Math Processing Error`, which will already be in the error array.  I can add an additional message, if you think that is a help, but there should already be an error there already.  Let me know if you think an additional message is necessary and I'll modify the PR.

PR #451, opened 2019-09-05
Pin Mathjax version to 2.7.6

matthewborden:
Mathjax just updated to version 3.


dpvc:
Thanks for your interest in MathJax and mathjax-node; we appreciate the support.  There are some problems with your pull request, however.

First, it is based on a branch that is over two years old (July 2017).  Small changes like this should be based on a recent branch.  Also, it should be branched from and targeted to the develop branch, not the master branch.

Second, this branch includes commits for additional functionality (the macro reset functionality), which makes this PR effectively the same as PR #348.  Your PR should include only the single fix that you have indicated in the title.

Third, the issue with the MathJax version was resolved in commit 3418e25, which set the version to `^2.7.2`, so it will get the highest 2.x version, but not 3.0.  This was made in Dec 2017, so that is nearly two years ago.

Finally, there are merge conflicts with the branch (see for example the `package.json` file, where there are conflicts due to the Dec 2017 change and later ones).

For these reasons, this PR will be rejected.  The issue you are indicating in your title is already taken care of, so there is no need for a new PR for this.

